PR #17, opened 2024-04-30
Added simdjson dependency

AndreyPavlenko:


kurapov-peter:
@AndreyPavlenko, onednn doesn't use simdjson, why do we need it?

AndreyPavlenko:
It uses an internal parser implementation.

PR #18, opened 2024-04-30
Add CI

leshikus:
Using self-hosted runner requires rebuilding the corresponding docker image, it lacks `cmake` as for now


kurapov-peter (approved):
LGTM

PR #32, opened 2024-05-02
oneDNN interface implementation layout

AndreyPavlenko:


tests/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+endforeach ()
+
+file(GLOB TEST_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/test_*.cpp)
+
+foreach (TEST_SOURCE ${TEST_SOURCES})
+    get_filename_component(TEST_NAME ${TEST_SOURCE} NAME_WE)
+    add_executable(${TEST_NAME} ${TEST_SOURCE})
+    target_link_libraries(${TEST_NAME} graph_compiler_static)
+    add_test(NAME ${TEST_NAME}
+            COMMAND ${TEST_NAME}
```
kurapov-peter, line 18: How would we handle a test with additional options?

tests/test_dnnl_exec_json.cpp
kurapov-peter, line 1: This should go through gtest

lib/CMakeLists.txt
```diff
@@ -1 +1,17 @@
-add_subdirectory(gc-dialects)
\ No newline at end of file
+add_subdirectory(gc-dialects)
+
+file(GLOB SOURCES "*.*pp")
```
kurapov-peter, line 3: This looks error-prone.

lib/CMakeLists.txt
```diff
@@ -1 +1,17 @@
...
+add_library(graph_compiler SHARED
+        ${SOURCES}
+)
+add_library(graph_compiler_static STATIC
+        ${SOURCES}
+)
```
kurapov-peter, line 9: What are we building the two libraries for?

lib/dnnl_graph_compiler.cpp
```diff
@@ -0,0 +1,54 @@
...
+// dnnl_graph_compiler.h interface implementation.
+// TODO: Implement
+
+using namespace gc;
+
+struct dnnl_graph_compiler {
+  const graph_compiler gc;
+
+  explicit dnnl_graph_compiler(const struct dnnl_graph_compiler_context *ctx)
+      : gc(ctx->num_threads, ctx->allocator, ctx->deallocator) {}
```
kurapov-peter, line 13: what happens if ctx is `nullptr`?

lib/dnnl_graph_compiler.cpp
```diff
@@ -0,0 +1,54 @@
+#include "dnnl_graph_compiler.h"
```
kurapov-peter, line 1: Why are we calling it dnnl btw?

lib/dnnl_graph_compiler.cpp
kurapov-peter, line 1: I expected this would implement the interface. Why are we creating a wrapper?

kurapov-peter (changes requested):

lib/CMakeLists.txt
AndreyPavlenko: The static is used by tests, because in the shared lib only the public API functions are exposed.

lib/dnnl_graph_compiler.cpp
AndreyPavlenko: This is an implementation of the C interface dnnl_graph_compiler.h and i'ts a wrapper around the graph_compiler.hpp interface, which is the actual graph compiler implementation.

include/graph_compiler.hpp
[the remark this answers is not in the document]
AndreyPavlenko: Moved.

tests/CMakeLists.txt
AndreyPavlenko: To be added when required.

tests/test_dnnl_exec_json.cpp
AndreyPavlenko: Done

lib/dnnl_graph_compiler.cpp
AndreyPavlenko: Removed the wrapper.

cmake/functions.cmake
```diff
@@ -0,0 +1,92 @@
...
+    endif ()
+
+    if (NOT DEFINED ${name}_VERSION_COUNT)
+        if (GC_${uname}_SKIP_ADD)
+            if (NOT DEFINED ${name}_POPULATED)
+                FetchContent_Populate(${name})
+                FetchContent_GetProperties(${name})
+                set(${name}_POPULATED TRUE PARENT_SCOPE)
+                set(${name}_SOURCE_DIR ${${name}_SOURCE_DIR} PARENT_SCOPE)
+                set(${name}_BINARY_DIR ${${name}_BINARY_DIR} PARENT_SCOPE)
```
kurapov-peter, line 67: What's the difference between these and `FetchContent_MakeAvailable`?

cmake/onednn.cmake
```diff
@@ -1,27 +1,25 @@
...
-FetchContent_Populate(oneDNN)
-FetchContent_GetProperties(oneDNN)
-message(STATUS "oneDNN source dir: ${onednn_SOURCE_DIR}")
-FetchContent_MakeAvailable(oneDNN)
\ No newline at end of file
+include_directories(
+        ${dnnl_BINARY_DIR}/include
+        ${dnnl_SOURCE_DIR}/include
+        ${dnnl_SOURCE_DIR}/src/graph/backend/elyzor/include
+)
```
kurapov-peter, line 25: We should probably export some `${name}_INCLUDES` and then use it properly in `target_include_directories`.

cmake/functions.cmake
```diff
@@ -0,0 +1,92 @@
...
+            FetchContent_MakeAvailable(${name})
+        endif ()
+    endif ()
+endfunction()
+
+# Add sources to the global list GC_LIB_SOURCES.
+# The sources could be specified as a list of files or a GLOB pattern:
+#   gc_add_lib_sources(GLOB "src/*.cpp")
+#   gc_add_lib_sources("src/foo.cpp" "src/bar.cpp")
+function(gc_add_lib_sources sources)
```
kurapov-peter, line 79: So far we have a huge function that adds two source files to a single library. Do we really need it?

src/dnnl/dnnl_graph_compiler.hpp
```diff
@@ -0,0 +1,25 @@
...
+struct dnnl_graph_compiler {
+  const dnnl_graph_compiler_context ctx;
+
+  explicit dnnl_graph_compiler(const dnnl_graph_compiler_context *context)
+      : ctx(*context) {}
+
+  const dnnl_graph_compiler_executable *
+  compile(const std::string_view &graph_json) const;
+
+  void execute(const struct dnnl_graph_compiler_executable *exe,
```
kurapov-peter, line 20: `execute` does not belong to the compiler.

src/dnnl/dnnl_graph_compiler.hpp
```diff
@@ -0,0 +1,25 @@
...
+  // TODO: Implement
+};
+
+struct dnnl_graph_compiler {
+  const dnnl_graph_compiler_context ctx;
+
+  explicit dnnl_graph_compiler(const dnnl_graph_compiler_context *context)
+      : ctx(*context) {}
+
+  const dnnl_graph_compiler_executable *
```
kurapov-peter, line 17: Can it be a `unique_ptr`?

kurapov-peter (changes requested):
Looks better now, a couple of tweaks

cmake/functions.cmake
AndreyPavlenko: FetchContent_MakeAvailable() adds the populated content into the build.

cmake/functions.cmake
AndreyPavlenko: For only 2 sources - definitely not. But I think there will be more sources and more subdirectories.  The idea is to use this function in subdirectories to add sources to the lib.

cmake/onednn.cmake
AndreyPavlenko: Now  the script sets the DNNL_INCLUDES variable.

src/dnnl/dnnl_graph_compiler.hpp
AndreyPavlenko: Moved to the executable class.

src/dnnl/dnnl_graph_compiler.hpp
AndreyPavlenko: Done.

test/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+if (NOT GC_TEST_ENABLE)
+    message(STATUS "The tests are not enabled.")
+    return()
+endif ()
+
+include(gtest)
+
+# Static library to be used in tests
+add_library(graph_compiler_static STATIC ${GC_LIB_SOURCES})
```
Menooker, line 9: A question regarding gtest: is it compatible with LLVM's lit+gtest-unittest? We may need some C++-based unittests for MLIR in the future. They may use `add_mlir_unittest` in cmake which is provided in MLIR upstream, to register a unittest in the llvm-lit system. `add_mlir_unittest` will make the target depend on LLVM's copy of gtest.

Devjiu: definition of "add_mlir_unittest" seems trivial:
```cmake
function(add_mlir_unittest test_dirname)
  add_unittest(MLIRUnitTests ${test_dirname} ${ARGN})
endfunction()
```
And as far as I know lit tests and untitests (aka gtest) are not connected. Both of them used in llvm/mlir and unittests located in some folder like: `llvm-project/mlir/unittests/Analysis/Presburger`, but lit is `llvm-project/mlir/test/Transforms/control-flow-sink.mlir` and they have different cmakes and so on. 
What do you mean by lit+gtest?

Menooker: > definition of "add_mlir_unittest" seems trivial:

Please look deeper into the definition of `add_unittest`: it is not provided by cmake. Instead, it is defined in AddLLVM.cmake of upstream LLVM. It sets some path like the config path for `lit`.

> And as far as I know lit tests and untitests (aka gtest) are not connected

No, it's not true. Please try the cmake target `check-mlir` or  `check-mlir-unit` in upstream LLVM/MLIR. It will build and run the unittest (in the context of MLIR, they call gtest "unittest"). I think `llvm-lit` is the top-level test driver for all of `FileCheck` `Python` and `gtests`.
I did some research on how unittest is linked with lit. Besides creating executables via `add_unittest`, MLIR upstream adds the unittest cmake-build target to `MLIR_TEST_DEPENDS` at `mlir/test/CMakeLists.txt`. The targets in `MLIR_TEST_DEPENDS` will be marked as dependency at `add_lit_testsuite` below.

Also, MLIR upstream has a directory at `mlir/test/Unit`. The `lit.cfg.py` in it indicates that it is a GoogleTest suite. "add_mlir_unittest" seems to internally adjust the output directory to make sure the cfg is generated at the right place.

Devjiu: Okay, now I get your point. I've added lit tests in the similar way with tpp repo - https://github.com/intel/graph-compiler/pull/33. 

I guess for the first period of time we will just use lit and run gtests separately. I think we should understand how many mlir/llvm unittests we need and if there are will be many of them we should try to use this approach. 

I agree with you, but we should back to this when we will have understanding of which tests we need.

PR #34, opened 2024-05-03
Add googletest dep

kurapov-peter:


kurapov-peter:
Will be merged as a part of #32

PR #35, opened 2024-05-06
build: add llvm_update_compile_flags and mlir_check_all_link_libraries

Menooker:
Add calls to `llvm_update_compile_flags` and `mlir_check_all_link_libraries` for `gc-opt`, as MLIR upstream does for mlir's own executable targets.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -3,5 +3,7 @@ get_property(conversion_libs GLOBAL PROPERTY MLIR_CONVERSION_LIBS)
 set(gc_opt_libs ${dialect_libs} ${conversion_libs} MLIROptLib)
 add_llvm_executable(gc-opt gc-opt.cpp)
 target_link_libraries(gc-opt PRIVATE ${gc_opt_libs})
+llvm_update_compile_flags(gc-opt)
+mlir_check_all_link_libraries(gc-opt)
 
 add_subdirectory(dnnl)


src/CMakeLists.txt
```diff
@@ -2,4 +2,4 @@ get_property(dialect_libs GLOBAL PROPERTY MLIR_DIALECT_LIBS)
 get_property(conversion_libs GLOBAL PROPERTY MLIR_CONVERSION_LIBS)
 set(gc_opt_libs ${dialect_libs} ${conversion_libs} MLIROptLib)
 add_llvm_executable(gc-opt gc-opt.cpp)
-target_link_libraries(gc-opt ${gc_opt_libs})
\ No newline at end of file
+target_link_libraries(gc-opt PRIVATE ${gc_opt_libs})
```
Menooker, line 5: I have a question: MLIR's own tools also addes

```
llvm_update_compile_flags(mlir-cat)
mlir_check_all_link_libraries(mlir-cat)
```

after `target_link_libraries`. Shall we also add the code?

kurapov-peter:
I believe this is fixed in #32

src/CMakeLists.txt
kurapov-peter: Sure, go for it!

Menooker: OK. I have updated the PR to add `llvm_update_compile_flags` and `mlir_check_all_link_libraries` only.

Menooker:
Hi @kurapov-peter, I am curious why we have both `lib` and `src` directories, and what are they for, repectively? In the future, we may need our own mlir-cpu-runner and our runtime support library, where do you suggest to put them in?

I have another question: shall we split the code for `src/gc-opt.cpp` and `src/dnnl/*` in different directories? Currently they share the same parent CMakelist.txt, and changes for gc-opt may affect the dnnl subdirectory.

kurapov-peter (approved):

kurapov-peter:
> I am curious why we have both lib and src directories, and what are they for, repectively?

No particular reason. Lib follows the MLIR convention. Src is for everything else that is not a dialect/pass/whatever.

> In the future, we may need our own mlir-cpu-runner and our runtime support library, where do you suggest to put them in?

We definitely need one. Src or a separate folder.

> shall we split the code for src/gc-opt.cpp and src/dnnl/* in different directories?

So far it's simple enough not to care too much about it. Let's split when it becomes complex.

PR #36, opened 2024-05-06
Add no-unused-but-set-parameter to skip verbose warning

Menooker:
I received the following warning messages many times while compiling. g++ version 9.4.0

```
/home/shared/bin/mlir/20240505/include/mlir/IR/Attributes.h: In instantiation of ‘static bool llvm::CastInfo<To, From, typename std::enable_if<(is_same_v<mlir::Attribute, typename std::remove_const<From>::type> || is_base_of_v<mlir::Attribute, From>), void>::type>::isPossible(mlir::Attribute) [with To = mlir::Attribute; From = mlir::Attribute]’:
/home/shared/bin/mlir/20240505/include/llvm/Support/Casting.h:311:29:   required from ‘static To llvm::DefaultDoCastIfPossible<To, From, Derived>::doCastIfPossible(From) [with To = mlir::Attribute; From = mlir::Attribute; Derived = llvm::CastInfo<mlir::Attribute, mlir::Attribute, void>]’
/home/shared/bin/mlir/20240505/include/llvm/Support/Casting.h:657:46:   required from ‘decltype(auto) llvm::dyn_cast(From&) [with To = mlir::Attribute; From = mlir::Attribute]’
/home/shared/bin/mlir/20240505/include/mlir/IR/OpImplementation.h:972:38:   required from ‘mlir::ParseResult mlir::AsmParser::parseAttribute(AttrType&, mlir::Type, llvm::StringRef, mlir::NamedAttrList&) [with AttrType = mlir::Attribute]’
/home/shared/bin/mlir/20240505/include/mlir/IR/OpImplementation.h:948:58:   required from here
/home/shared/bin/mlir/20240505/include/mlir/IR/Attributes.h:417:49: warning: parameter ‘ty’ set but not used [-Wunused-but-set-parameter]
  417 |   static inline bool isPossible(mlir::Attribute ty) {
```

This PR disables the warnings for non-MSVC builds.


kurapov-peter:
@Menooker, is it for MLIR build? If so, I'd suggest we set the flag for the target only and use stricter rules for our code.

Menooker:
> @Menooker, is it for MLIR build? If so, I'd suggest we set the flag for the target only and use stricter rules for our code.

Yes, it is for MLIR only. I have updated the code to let the CXX flag only affect on MLIR targets only.

CMakeLists.txt
```diff
@@ -41,6 +41,16 @@ set(GC_LIB_SOURCES CACHE INTERNAL "The graph_compiler library source paths")
 set(GC_LIB_INCLUDES CACHE INTERNAL "The graph_compiler library include paths")
 
 add_definitions(${LLVM_DEFINITIONS})
+
+set(GC_MLIR_CXX_FLAGS "")
+# Silence a false positive GCC -Wunused-but-set-parameter warning in constexpr
+# cases, by marking SelectedCase as used. See
+# https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85827 for details. The issue is
+# fixed in GCC 10.
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "10.0")
```
Menooker, line 50: The lines are copied from MLIR upstream. They have the same issue too.

ZhennanQin (approved):

PR #39, opened 2024-05-06
Copy llvm-lit to the llvm installation

kurapov-peter:
A bit hacky, but lits are not a part of the llvm installation in any way. We could tar the build itself, but it's larger. Thoughts?


leshikus (approved):
LGTM

Devjiu (approved):

PR #44, opened 2024-05-08
[mlir][Memref] Add memref-merge optimization

Menooker:


Menooker:
Design doc at `https://github.com/intel/graph-compiler/blob/45a02a4aafef315b65e5d5bfdc53b6ad2a062f06/docs/memref_schedule.md`

Menooker:
@Devjiu Here in this PR, I have added a unittest in lit. It requires LLVM to be built with `LLVM_INSTALL_GTEST=ON`

Devjiu:
> @Devjiu Here in this PR, I have added a unittest in lit. It requires LLVM to be built with `LLVM_INSTALL_GTEST=ON`

I like approach, but pr is a quite big. To change LLVM build options you need to update: .github/workflows/build-llvm.yml as it's done in #39. 
Also in doc you are sharing result of bufferization - It's a oneshot I guess. 
Maybe it's bettter to have test - that applies bufferization to your original mlp example. 
```bash
func.func @mlp(%x: tensor<128x128xf32>, %y: tensor<128x128xf32>) -> tensor<128x128xf32> {
   %a0 = tensor.empty() : tensor<128x128xf32>
   %a = linalg.matmul ins(%x, %y: tensor<128x128xf32>, tensor<128x128xf32>) outs(%a0: tensor<128x128xf32>) -> tensor<128x128xf32>
   %b0 = tensor.empty() : tensor<128x128xf32>
   %b = linalg.matmul ins(%a, %y: tensor<128x128xf32>, tensor<128x128xf32>) outs(%b0: tensor<128x128xf32>) -> tensor<128x128xf32>
   %c0 = tensor.empty() : tensor<128x128xf32>
   %c = linalg.matmul ins(%b, %y: tensor<128x128xf32>, tensor<128x128xf32>) outs(%c0: tensor<128x128xf32>) -> tensor<128x128xf32>
   %d0 = tensor.empty() : tensor<128x128xf32>
   %d = linalg.matmul ins(%c, %y: tensor<128x128xf32>, tensor<128x128xf32>) outs(%d0: tensor<128x128xf32>) -> tensor<128x128xf32>
   return %d : tensor<128x128xf32>
}
```
in my case `./build/src/gc-opt --one-shot-bufferize  test/gc-dialects/Linlagx/mlp.mlir` : 
```bash
module {
  func.func @mlp(%arg0: tensor<128x128xf32>, %arg1: tensor<128x128xf32>) -> tensor<128x128xf32> {
    %0 = bufferization.to_memref %arg1 : memref<128x128xf32, strided<[?, ?], offset: ?>>
    %1 = bufferization.to_memref %arg1 : memref<128x128xf32, strided<[?, ?], offset: ?>>
    %2 = bufferization.to_memref %arg1 : memref<128x128xf32, strided<[?, ?], offset: ?>>
    %3 = bufferization.to_memref %arg1 : memref<128x128xf32, strided<[?, ?], offset: ?>>
    %4 = bufferization.to_memref %arg0 : memref<128x128xf32, strided<[?, ?], offset: ?>>
    %alloc = memref.alloc() {alignment = 64 : i64} : memref<128x128xf32>
    linalg.matmul ins(%4, %3 : memref<128x128xf32, strided<[?, ?], offset: ?>>, memref<128x128xf32, strided<[?, ?], offset: ?>>) outs(%alloc : memref<128x128xf32>)
    %alloc_0 = memref.alloc() {alignment = 64 : i64} : memref<128x128xf32>
    linalg.matmul ins(%alloc, %2 : memref<128x128xf32>, memref<128x128xf32, strided<[?, ?], offset: ?>>) outs(%alloc_0 : memref<128x128xf32>)
    %alloc_1 = memref.alloc() {alignment = 64 : i64} : memref<128x128xf32>
    linalg.matmul ins(%alloc_0, %1 : memref<128x128xf32>, memref<128x128xf32, strided<[?, ?], offset: ?>>) outs(%alloc_1 : memref<128x128xf32>)
    %alloc_2 = memref.alloc() {alignment = 64 : i64} : memref<128x128xf32>
    linalg.matmul ins(%alloc_1, %0 : memref<128x128xf32>, memref<128x128xf32, strided<[?, ?], offset: ?>>) outs(%alloc_2 : memref<128x128xf32>)
    %5 = bufferization.to_tensor %alloc_2 : memref<128x128xf32>
    return %5 : tensor<128x128xf32>
  }
}
```

Also in MLIR there are specific methods `getUses()` and `getUsers()` that can be used for live ranges in the same way as you do with ticks. 
In MLIR there are liveness analysis  https://mlir.llvm.org/doxygen/Liveness_8h_source.html - can we reuse it? Your approach is fine, but have you tried to use existing mlir tools?

kurapov-peter (commented):
A couple of things: I'd suggest creating a similar RFC in the upstream to get the initial feedback. As to the PR itself, I'd split the testing and the pass implementation, just for the sake of reviews.

Menooker:
> In MLIR there are liveness analysis

Thanks for pointing out. I had a brief look into that and I find that MLIR's liveness analysis does not meet our requirements.
 * we need special handling for complex control flow, e.g., to make sure the buffers used in a for-loop has overlapping liveness. MLIR's liveness analysis simply checks the Block-Value-Operation relations
 * Our compile-time memory allocator needs a serialized stream of alloc/dealloc "events". We need to serialize the liveness info from analysis
 * We need to track the memref.view on memref.alloc.

I believe we can indeed build our analysis on MLIR's liveness analysis, but I doubt whether it can reduce the code complexity, since we have some non-standard requirement on it.

Menooker:
> As to the PR itself, I'd split the testing and the pass implementation

Did you mean splitting the the "adding our first unittest to lit" from this PR, or splitting all the testing code from the implementation?

Menooker:
@Devjiu @kurapov-peter I have opened a new PR https://github.com/intel/graph-compiler/pull/50 for updating the build system to enable unittest in MLIR-related code. Shall we move the discussion on testing to that PR?

Menooker:
> Maybe it's bettter to have test - that applies bufferization to your original mlp example.

It is good advice! I missed that part of the test and after adding that, it does exposes a bug. Thanks!

I have now added a `.mlir` test case for tensor->bufferization->merge-alloc pipeline. It now works as expected.

lib/gc-dialects/Transforms/MergeAlloc.cpp
```diff
@@ -0,0 +1,327 @@
...
+    }
+    if (!isa<scf::ForOp>(parent)) {
+      return parent;
+    }
+  }
+}
+
+FailureOr<size_t> getAllocSize(Operation *op) {
+  auto refType = op->getResultTypes().front().cast<MemRefType>();
+  int64_t size = refType.getElementTypeBitWidth() / 8;
```
ZhennanQin, line 97: Need to check whether it works for `i1` as it's a boolean type.

Menooker:
> can we discuss this approach a little bit more?
As told Petr this pass looks generic enough to be upstreamed.

Sure. We can directly discusses on teams if you wish to. :)

> e.g. in bufferization there are pass "OwnershipBasedBufferDeallocation"("ownership-based-buffer-deallocation")
and in it's implementation

Sorry, I don't quite understand how it is related to our pass. "OwnershipBasedBufferDeallocation" doesn't need liveness analysis (in the code you posted). I totally agree that we should use the components already in MLIR. Let's find which one we can reuse in our pass.

> BTW I don't fully understand purpose of "test.source" op?

It is a general operation that references a memref. In MLIR, you can write IR code with undefined dialects. And the passes can still work on these undefined operations, as long as we don't need the semantics of this operation. You can view "test.source" op as any operations referencing memref, like `memref.load`, `linalg.matmul`, etc.

Menooker:
I have submitted a RFC in the community: https://discourse.llvm.org/t/rfc-compile-time-memref-alloc-scheduling-merging-optimization/78872

ciyongch:
> can we discuss this approach a little bit more? As told Petr this pass looks generic enough to be upstreamed.
> 
> e.g. in bufferization there are pass "OwnershipBasedBufferDeallocation"("ownership-based-buffer-deallocation") and in it's implementation

Hi @Devjiu , this pass is designed to handle `memref` and thus run after the bufferization pass,  the bufferization pass itself doesn't do such kind of memory optimization, and the community is still in an active transition from conversion-based to one-shot-bufferization(DPS style). The current algorithm is only verified on CPU platform in Graph compiler v1, it might require broader coverage for upstream. I would suggest we move on with this implementation and keep track with upstream feedback to catch up our internal timeline.

lib/gc/Transforms/MergeAlloc.cpp
```diff
@@ -0,0 +1,352 @@
...
+      return;
+    }
+    std::unordered_map<uintptr_t, std::vector<uintptr_t>> dummy;
+    for (auto &[scope, traces] : *tracesOrFail) {
+      std::unordered_map<uintptr_t, std::size_t> outSchedule;
+      if (traces.empty())
+        continue;
+      auto total = mlir::gc::memoryplan::schedule_memory_allocations(
+          traces, 64, !this->optionNoLocality,
+          gc::memoryplan::inplace_info_map(), outSchedule, dummy);
```
AIYoungcino, line 321: Excuse me, may I ask - can “gc::memoryplan::inplace_info_map()” be automatically derived through static analysis?

Menooker: The memory allocation planner of current PR is extracted from our legacy code, which supports in-place operations. Currently, we don't support in-place allocation in the current design, but I keep the interface so that we may enable it in the future. By in-place operations, I mean operations like `Z = linalg.add X, Y`, where `X` can reuse buffer of `Z`. However, in the current implementation, `X, Y, Z` have overlapping lifetime.

ciyongch:
We might need an issue to track this as well.

Menooker:
Updated the code. Ready for review.

include/gc/Transforms/StaticMemoryPlanning.h
```diff
@@ -0,0 +1,74 @@
...
+  // unique id of a buffer
+  uintptr_t bufferId;
+  // if > 0, size of the buffer allocation, if = 0, it is a deallocation trace
+  std::size_t size;
+  MemoryTrace(uintptr_t bufferId = 0, std::size_t size = 0)
+      : bufferId{bufferId}, size{size} {}
+};
+
+using ScopeTraceData =
+    llvm::DenseMap<Operation *, llvm::SmallVector<memoryplan::MemoryTrace, 8>>;
```
ciyongch, line 41: Seems it's useless?

lib/gc/Transforms/StaticMemoryPlanning.cpp
```diff
@@ -0,0 +1,657 @@
...
+
+// the memory chunk that is split from another chunk
+struct split_chunk_t : public MemoryChunk {
+  MemoryChunk *parent;
+  // if the chunk is the left hand side (smaller starting offset)
+  bool is_lhs_;
+  split_chunk_t(size_t size, MemoryChunk *parent, bool is_lhs)
+      : MemoryChunk{ChunkType::SPLIT, size}, parent(parent), is_lhs_(is_lhs) {}
+  void move(int64_t startDiff) override {
+    if (is_lhs_) {
```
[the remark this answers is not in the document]
Menooker: Currently, we only support splitting into 2 pieces.

lib/gc/Transforms/StaticMemoryPlanning.cpp
```diff
@@ -0,0 +1,657 @@
...
+  size_t start;
+  OriginChunk(size_t start, size_t size)
+      : MemoryChunk{ChunkType::ORIGIN, size}, start(start) {}
+  void move(int64_t startDiff) override { start += startDiff; }
+  void extend(int64_t sizeDiff) override { size += sizeDiff; }
+  size_t getStartOffsetImpl() override { return start; }
+};
+
+// the memory chunk that is split from another chunk
+struct split_chunk_t : public MemoryChunk {
```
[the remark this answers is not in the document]
Menooker: Sorry, I missed that. There were too many of them for code-style changes. :)

ciyongch:
to address the issue of https://github.com/intel/graph-compiler/issues/269

ciyongch:
Please rebase the PR, and let's have a test with MLP to see the perf gain.

ciyongch (approved):

ciyongch:
@yifeizh2 can you help to pull in this patch into MLP pattern to see if any performance gain?

yifeizh2:
> @yifeizh2 can you help to pull in this patch into MLP pattern to see if any performance gain?

It has performance gain under most of the cases. I will double check the IR later to make confirmation

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/yifeizh2/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/yifeizh2/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">


dtype | batch size | hidden list | w/o mem merge | w/ mem merge | Ratio
-- | -- | -- | -- | -- | --
bf16 | 128 | 16x512 |   |   |  
bf16 | 128 | 512x256 | 0.0527 | 0.0514 | 102.50%
bf16 | 128 | 256x128 | 0.0354 | 0.0384 | 92.16%
bf16 | 128 | 512x1024 | 0.0501 | 0.0362 | 138.19%
bf16 | 128 | 1024x1024 | 0.1004 | 0.0899 | 111.70%
bf16 | 128 | 1024x512 | 0.1041 | 0.0950 | 109.53%
bf16 | 128 | 512x256 | 0.0519 | 0.0518 | 100.30%
bf16 | 128 | 16x512x256x128 |   |   |  
bf16 | 128 | 512x1024x1024x512x256 | 0.3246 | 0.3024 | 107.33%



</body>

</html>

ciyongch:
@kurapov-peter @ZhennanQin @zhczhong @Yun-Fly Please help to review and approve if no other comments :)

include/gc/Transforms/Passes.h
```diff
@@ -42,9 +50,68 @@ class XeGPUDialect;
...
+  int64_t alignment = 64;
+  TraceCollectorFunc tracer;
+  MemoryPlannerFunc planner;
+  MemoryMergeMutatorFunc mutator;
+};
+
+/// Creates an operation pass to merge the local memref allocations
+std::unique_ptr<Pass> createMergeAllocPass(const MergeAllocationOptions &o);
+std::unique_ptr<Pass> createMergeAllocPass();
+
```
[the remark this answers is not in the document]
Yun-Fly: BTW, it seems that this pass not be added into our pipeline?

lib/gc/Transforms/StaticMemoryPlanning.cpp
```diff
@@ -0,0 +1,656 @@
...
+      cur_offset += buf->size;
+    }
+    return ss.str();
+  }
+};
+
+size_t scheduleMemoryAllocations(
+    const Traces &traces, std::size_t alignment, bool hotFirst,
+    const InplaceInfoMap &inplaceMap,
+    std::unordered_map<uintptr_t, std::size_t> &outSchedule,
```
Yun-Fly, line 631: One open here: is there any possible to estimate allocation size during fusion stage, just like what we do in v1?

lib/gc/Transforms/MergeAllocTickBased.cpp
```diff
@@ -0,0 +1,423 @@
...
+      auto allocscope = getAllocScope(&s, alloc);
+      alloc->setAttr(
+          "__mergealloc_lifetime",
+          DenseI64ArrayAttr::get(root->getContext(),
+                                 {reinterpret_cast<int64_t>(allocscope),
+                                  tick.firstAccess, tick.lastAccess}));
+      allocscope->setAttr(
+          "__mergealloc_scope",
+          IntegerAttr::get(mlir::IntegerType::get(root->getContext(), 64),
+                           reinterpret_cast<int64_t>(allocscope)));
```
Yun-Fly, line 337: Better to use `constexpr StringLiteral mergeAllocXXXAttrName = "__mergealloc_xxx__"` instead of raw string, similar to upstream `__inplace_operands_attr__`, which would be friendly if we want to get attribute in the future.

include/gc/Transforms/Passes.h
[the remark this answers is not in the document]
Menooker: We require a special argument for pass configuration and it cannot be set by simple pass-args. So we need to write the pass creators by our own.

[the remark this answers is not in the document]
Menooker: The pass needs further development of @ciyongch to handle nested parallel. It may be better to add the pass at his enhancement is ready?

lib/gc/Transforms/StaticMemoryPlanning.cpp
Menooker: Yes. The internals of the pass is already described at the doc. You can skip the IR mutation stage and get the mem scheduling results.

ZhennanQin (approved):

lib/gc/Transforms/MergeAllocTickBased.cpp
Menooker: It is for testing only. The data is not actually passed via attr.

lib/gc/Transforms/StaticMemoryPlanning.cpp
Yun-Fly: After offline sync, this pass expects `memref` rather than `tensor` dialect. So, it needs more changes to be applied during fusion stage.

Yun-Fly (approved):

PR #47, opened 2024-05-08
[LIT] Fix lit copying path.

Devjiu:
This commit changes directory which used to copy llvm-lit binary.


Devjiu:
Waiting for https://github.com/intel/graph-compiler/actions/runs/9002613715 it breaks all builds at the current moment.

Devjiu:
suppressed by #46

PR #50, opened 2024-05-09
[tests] Enable MLIR C++ unittest with lit

Menooker:
Add an example gtest unittest for MLIR and let it be driven by lit. The required cmake and lit configs are also added.


Menooker:
This PR requires a new build of LLVM `-DLLVM_INSTALL_GTEST=ON`, otherwise it will fail to find gtest header and static library. How do I let CI re-build LLVM?

ZhennanQin (approved):

cmake/llvm-version.txt
```diff
@@ -1 +1 @@
-179e174945b6c0da462c534504720c9544aebf84
+5adcfd4c17826b2b8f023881baa1c7f79cb23920
```
ciyongch, line 1: It should be fine to combine the llvm version update and other changes in a single PR at this stage, but it'd be better to split this in future?

test/gc-dialects/Unit/lit.cfg.py
```diff
@@ -0,0 +1,50 @@
+# -*- Python -*-
```
ciyongch, line 1: Maybe we don't need the multiple directories like `gc-dialects` to share the `lit` python files for other components like transforms/passes, runtime and others?

Menooker: The C++-based unittest is a special case. Filecheck tests don't need this lit config file.

cmake/llvm-version.txt
Menooker: It is to trigger LLVM rebuilt in CI. We need a new flag in cmake options of LLVM.

ciyongch (approved):

PR #65, opened 2024-05-11
add cpu runner

Menooker:
Add `gc-cpu-runner`. Also move `gc-opt` in `src/gc-opt` directory. LLVM's cmake function forces us to do that. If we add more than 1 target in the same directory, it complains:

```
CMake Error at /llvm/lib/cmake/llvm/LLVMProcessSources.cmake:116 (message):
  Found erroneous configuration for source file gc-cpu-runner.cpp

  LLVM's build system enforces that all source files are added to a build
  target, that exactly one build target exists in each directory, and that
  this target lists all files in that directory.  If you want multiple
  targets in the same directory, add PARTIAL_SOURCES_INTENDED to the target
  specification, though it is discouraged.
```

So we cannot put `gc-cpu-runner.cpp` and `gc-opt.cpp` in the same directory.

Another change to the build system is that we change the output LLVM tool binary directory to `bin/` of the cmake build directory, to make sure `gc-cpu-runner` and `gc-opt` are generated in the same directory, for the sake of `lit` to correctly find the binaries.


src/gc-cpu-runner/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+  if(NOT LLVM_LINK_LLVM_DYLIB)
+    set(LLVM_LINK_COMPONENTS
+      Core
+      Support
+      nativecodegen
+      native
+    )
+  endif()
+  set(MLIR_LINK_COMPONENTS
+    MLIRAnalysis
```
ZhennanQin, line 20: Please add a comment about where to get this list.

Menooker: OK, done.

ZhennanQin (approved):

kurapov-peter (approved):

ciyongch (approved):

PR #66, opened 2024-05-13
Enable dynamic linking MLIR and LLVM

Menooker:
Added a cmake option `GC_DEV_LINK_DYN_LLVM` for developer.  It reduces binary sizes of `gc-cpu-runner` and `gc-opt` from `36MB` and `85MB` to `0.2MB` and `2MB`. This helps to reduce the link time and the loading time of debuggers like `gdb`. This option is by default off.


Menooker:
depends on https://github.com/intel/graph-compiler/pull/65

CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
 option(GC_LEGACY_ENABLE ON)
 option(GC_TEST_ENABLE "Build the tests" ON)
+option(GC_DEV_LINK_DYN_LLVM "Link dynamic libraries of LLVM and MLIR. For developers only. Do not use it in packing the library." OFF)
```
ciyongch, line 13: Suggest to change the option name to `GC_DEV_LINK_LLVM_DYLIB` which is more align with LLVM's `LLVM_LINK_LLVM_DYLIB`?

Menooker: I agree with you. I will change that.

kurapov-peter (approved):
Looks good, just on minor comment

ciyongch (approved):

lib/gc/CMakeLists.txt
```diff
@@ -2,5 +2,7 @@ if(GC_MLIR_CXX_FLAGS)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${GC_MLIR_CXX_FLAGS}")
 endif()
 
+include(../../cmake/functions.cmake)
```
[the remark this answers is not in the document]
Menooker: After removing it, cmake complains missing command. Let's keep it for now.

Menooker:
Updates: after updating LLVM to a recent version, `gc-cpu-runner`'s binary size jumps to from `0.2MB` to `14MB`. Not sure why... Cmake internally uses `gc-sections` to track function-level dependencies. I guess some changes in upstream creates a dependency on all dialects in `JitRunner` module.

PR #74, opened 2024-05-15
[Transforms] Add constant_weights_folding pass

niuxiaog:
This PR implements constant_weights_folding pass:
- The input MLIR function will be split into two functions: compute() and fold(). The fold() function contains the constant operations whose input tensors and output tensors are all constant. The compute() function contains other operations that depend on the variable values and folded constant tensors.
- A constant tensor may be fully folded by several sequential operations. If fully folding increases the data size dramatically (i.e., by BroadcastOp), we choose a partial folding.
- Swaps the constant BroadcastOp and the constant operations after it, so that these constant operations can be included into partial folding. Currently there are strong constraints on these ops to ensure correctness.

A manager which allocates/manages/frees the buffers for folded constant weights is needed. It will be implemented later.

### DIFF
--- a/include/gc/Analysis/DataFlow/ConstantSubgraphAnalysis.h
+++ b/include/gc/Analysis/DataFlow/ConstantSubgraphAnalysis.h
@@ -1,0 +1,127 @@
+//===- ConstantSubgraphAnalysis.h - Constant subgraph analysis ------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements constant subgraph analysis. In this file are:
+// 1. the lattice value class that represents operations with constant inputs
+// and outputs in the program, and
+// 2. a sparse constant subgraph analysis.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef MLIR_ANALYSIS_DATAFLOW_CONSTANTSUBGRAPHANALYSIS_H
+#define MLIR_ANALYSIS_DATAFLOW_CONSTANTSUBGRAPHANALYSIS_H
+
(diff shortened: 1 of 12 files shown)


lib/gc/Transforms/ConstantSubgraphAnalysis.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+struct ConstantSubgraphAnalysis
+    : public impl::ConstantSubgraphAnalysisBase<ConstantSubgraphAnalysis> {
+  void runOnOperation() override;
+};
+
+void ConstantSubgraphAnalysis::runOnOperation() {
+  Operation *op = getOperation();
+  auto &func =
+      op->getRegions().front().getBlocks().front().getOperations().front();
```
ciyongch, line 38: Is there any shortcut for this kind of operation?

lib/gc/Transforms/Pipeline.cpp
```diff
@@ -53,6 +53,8 @@ void populateTensorPasses(mlir::OpPassManager &pm) {
   // todo: padding propagation pass
   // todo: layout propagation pass
   // todo: tensor constant propagation pass
+  pm.addPass(createConstantSubgraphAnalysisPass());
+  pm.addPass(createConstantTensorFoldingPass());
```
ciyongch, line 57: Shall we combine these two passes into one, and provide an option to do the analysis only is needed?

Menooker: I feel the same. Maybe we should put the analysis into the pass, unless the const-subgraph-analysis is needed by more than one pass.

niuxiaog: OK, I will put them into to one.

niuxiaog:
Though the PR is `ready_to_review`, we need PR #342 and `benchgc`'s support to fully enable this feature. For OpenVINO integration, this feature can be enabled by modifying [file](https://github.com/slyalin/openvino/pull/167/files#diff-1abc3d58e0f031e6a3e7e0a4ac6cc2faca3880e962980d87268275c1e331b615).

PR #79, opened 2024-05-16
Add kmp_* wrapper for gomp environment

Menooker:
omp dialect of MLIR is lowered to `kmp_*` function calls to libomp (and libiomp). These symbols are not provided by gomp. This will force the end-users to use clang instead of g++ to compile the runtime. To avoid this issue, as is recommended by the openmp team, we provide a wrapper over the standard omp primitives to implement the libomp symbols based on gomp. Currently limited features are supported, including static schedule omp-for and omp-section.

The wrapper will be disabled when the runtime is linked with libomp/libiomp.

### DIFF
--- a/lib/gc/ExecutionEngine/CPURuntime/CMakeLists.txt
+++ b/lib/gc/ExecutionEngine/CPURuntime/CMakeLists.txt
@@ -1,0 +1,15 @@
+find_package(OpenMP REQUIRED)
+
+if ("iomp" IN_LIST OpenMP_C_LIB_NAMES OR "omp" IN_LIST OpenMP_C_LIB_NAMES OR "omp5" IN_LIST OpenMP_C_LIB_NAMES)
+else()
+  add_definitions("-DGC_NEEDS_OMP_WRAPPER=1")
+endif()
+
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fopenmp")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fopenmp")
+add_mlir_library(GCCpuRuntime
+    SHARED
+    Parallel.cpp
+
+    EXCLUDE_FROM_LIBMLIR
+  )
--- a/lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
+++ b/lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
@@ -1,9 +1,9 @@
-//===-- Parallel.cpp - Definitions for parallel runtime -----------*- C++ -*-=//
-//
+//===-- Parallel.cpp - parallel ---------------------------------*- C++ -*-===//
+// 
 // This file is licensed under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-//
+// 
 //===----------------------------------------------------------------------===//
 
 #include <assert.h>
--- a/lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
+++ b/lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
@@ -74,6 +74,8 @@ int WEAK_SYMBOL __kmpc_global_thread_num(void *loc) {
   return omp_get_thread_num();
 }
 
+// The implementation was extracted and simplified from LLVM libomp
+// at openmp/runtime/src/kmp_sched.cpp
 void WEAK_SYMBOL __kmpc_for_static_init_8u(void *loc, int32_t gtid,
                                            int32_t schedtype,
                                            int32_t *plastiter, uint64_t *plower,
--- a/lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
+++ b/lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
@@ -1,9 +1,9 @@
 //===-- Parallel.cpp - parallel ---------------------------------*- C++ -*-===//
-// 
+//
 // This file is licensed under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// 
+//
 //===----------------------------------------------------------------------===//
 
 #include <assert.h>
--- a/lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
+++ b/lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
@@ -1,5 +1,4 @@
 //===- Parallel.cpp - Definitions for parallel runtime  -----------*- C++ -*-=//
-//-*-===//
 //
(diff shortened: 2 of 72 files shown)


lib/gc/ExecutionEngine/CPURuntime/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+find_package(OpenMP REQUIRED)
+
+if ("iomp" IN_LIST OpenMP_C_LIB_NAMES OR "omp" IN_LIST OpenMP_C_LIB_NAMES OR "omp5" IN_LIST OpenMP_C_LIB_NAMES)
```
ciyongch, line 3: Does this match "libiomp", "libiomp5", "libomp", "libomp5"?

lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
```diff
@@ -0,0 +1,189 @@
...
+} // namespace
+
+extern "C" {
+int gc_runtime_keep_alive = 0;
+void gc_arrive_at_barrier(barrier_t *b, barrier_idle_func idle_func,
+                          void *idle_args) {
+  auto cur_round = b->rounds_.load(std::memory_order_acquire);
+  auto cnt = --b->pending_;
+  assert(cnt >= 0);
+  //   int count = 0;
```
ciyongch, line 43: useless?

lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
```diff
@@ -0,0 +1,189 @@
...
+  if (cnt == 0) {
+    b->pending_.store(b->total_);
+    b->rounds_.store(cur_round + 1);
+  } else {
+    if (idle_func) {
+      if (cur_round != b->rounds_.load()) {
+        return;
+      }
+      idle_func(&b->rounds_, cur_round + 1, -1, idle_args);
```
ciyongch, line 52: why the `tid` is always passed as `-1`?

lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
```diff
@@ -0,0 +1,189 @@
...
+void WEAK_SYMBOL __kmpc_for_static_init_8u(void *loc, int32_t gtid,
+                                           int32_t schedtype,
+                                           int32_t *plastiter, uint64_t *plower,
+                                           uint64_t *pupper, int64_t *pstride,
+                                           int64_t incr, int64_t chunk) {
+  if (unlikely(schedtype != 34)) {
+    std::abort();
+  }
+  const int32_t FALSE = 0;
+  const int32_t TRUE = 0;
```
ciyongch, line 89: `TRUE = 1`?

lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
```diff
@@ -0,0 +1,189 @@
...
+  /* compute trip count */
+  if (incr == 1) {
+    trip_count = *pupper - *plower + 1;
+  } else if (incr == -1) {
+    trip_count = *plower - *pupper + 1;
+  } else if (incr > 0) {
+    // upper-lower can exceed the limit of signed type
+    trip_count = (UT)(*pupper - *plower) / incr + 1;
+  } else {
+    trip_count = (UT)(*plower - *pupper) / (-incr) + 1;
```
ciyongch, line 123: Seems we can combine all these into a single statement `trip_count = (UT)(*pupper - *plower) / incr + 1;`

lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
Menooker: We don't know our tid in this function, so just pass -1 instead.

lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
Menooker: It is copied and simplified from libomp of LLVM. I would like to keep the original code if possible.

lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
Menooker: It's a typo. Thanks for finding!

lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
Menooker: Removed now.

lib/gc/ExecutionEngine/CPURuntime/CMakeLists.txt
Menooker: Yes it should. :)

ciyongch: > Yes it should. :)

If so, then the `"omp" IN_LIST OpenMP_C_LIB_NAMES` will also match `libgomp`? And this is not what we want?

Menooker: > then the "omp" IN_LIST OpenMP_C_LIB_NAMES will also match libgomp

No, I suppose. libgomp has the name `gomp` in `OpenMP_C_LIB_NAMES`.  `IN_LIST` of cmake performs a full-name search, instead of finding substr.

> "libiomp", "libiomp5", "libomp", "libomp5"

It should only match `iomp omp omp5`

Menooker:
ping... any further comments?

ciyongch (approved):

ciyongch:
@ZhennanQin @kurapov-peter any comments from your side? We need to merge this as there's some PR dependency now.

lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
```diff
@@ -0,0 +1,188 @@
...
+};
+
+using barrier_idle_func = uint64_t (*)(std::atomic<int32_t> *remaining,
+                                       int32_t expected_remain, int32_t tid,
+                                       void *args);
+} // namespace
+
+extern "C" {
+int gc_runtime_keep_alive = 0;
+void gc_arrive_at_barrier(barrier_t *b, barrier_idle_func idle_func,
```
kurapov-peter, line 37: What's the usage scenario for the barrier?

ZhennanQin (approved):

lib/gc/ExecutionEngine/CPURuntime/Parallel.cpp
Menooker: it is not currently used for now. But we will introduce cpu barrier ops in CPURuntime in near future.

scripts/license.py
```diff
@@ -15,10 +15,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import datetime, sys, re, argparse
-from typing import Dict, Set
+from typing import Dict, Set, List
 
 WIDTH: int = 80
-intel_license: list[str] = [
+intel_license: List[str] = [
```
Menooker, line 21: @WangJialei-A Hi, in my local python (3.8.10), `list[str] ` is not allowed. I have updated the script here.

PR #81, opened 2024-05-16
[OneDNN Graph Dialect] Use Broadcast Trait and organize data types

BRUCE11111:
Some small changes：
1. Use broadcast trait to replace some functions.
2. Add some data types.

### DIFF
--- a/include/gc/Dialect/OneDNNGraph/OneDNNGraphTypes.td
+++ b/include/gc/Dialect/OneDNNGraph/OneDNNGraphTypes.td
@@ -17,14 +17,30 @@ include "OneDNNGraphDialect.td"
 // OneDNNGraph type definitions
 //===----------------------------------------------------------------------===//
 
+//===----------------------------------------------------------------------===//
+// Floating-point types.
+//===----------------------------------------------------------------------===//
+def OneDNNGraph_Float : AnyTypeOf<[F32,
+			                       F16,
+			                       BF16]>;
+
+//===----------------------------------------------------------------------===//
+// Integer types.
+//===----------------------------------------------------------------------===//
+
+def OneDNNGraph_Int8 : SI<8>;
+def OneDNNGraph_UInt8 : UI<8>;
+
+def OneDNNGraph_Int :AnyTypeOf<[OneDNNGraph_Int8,
+                                OneDNNGraph_UInt8
+                                ]>;
+
 def OneDNNGraph_DataType : AnyTypeOf<[
-                            F16,
-                            BF16, 
-                            F32, 
-                            SI<32>, 
-                            SI<8>, 
-                            UI<8>]>;
+                            OneDNNGraph_Float,
+                            OneDNNGraph_Int
+                            ]>;
 
 def OneDNNGraph_LogicalTensor : TensorOf<[OneDNNGraph_DataType]>;
+def OneDNNGraph_FloatTensor : TensorOf<[OneDNNGraph_Float]>;
 
 #endif // ONEDNNGRAPH_TYPES
--- a/include/gc/Dialect/OneDNNGraph/OneDNNGraphTypes.td
+++ b/include/gc/Dialect/OneDNNGraph/OneDNNGraphTypes.td
@@ -28,12 +28,8 @@ def OneDNNGraph_Float : AnyTypeOf<[F32,
 // Integer types.
 //===----------------------------------------------------------------------===//
 
-def OneDNNGraph_Int8 : SI<8>;
-def OneDNNGraph_UInt8 : UI<8>;
-
-def OneDNNGraph_Int :AnyTypeOf<[OneDNNGraph_Int8,
-                                OneDNNGraph_UInt8
-                                ]>;
+def OneDNNGraph_Int : AnyTypeOf<[SI<8>,
+                                 UI<8>]>;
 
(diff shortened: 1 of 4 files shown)


lib/gc/Dialect/OneDNNGraph/OneDNNGraphOps.cpp
```diff
@@ -65,11 +27,13 @@ LogicalResult onednn_graph::AddOp::inferReturnTypeComponents(
...
-  auto ret =
-      inferBroadcastShape<ValueShapeRange>(operands, outShape, getShapeIdx);
+  llvm::SmallVector<int64_t> input1, input2;
+  getShapeIdx(operands, 0).getDims(input1);
+  getShapeIdx(operands, 1).getDims(input2);
+  auto ret = OpTrait::util::getBroadcastedShape(input1, input2, outShape);
   inferredReturnShapes.push_back(
       ShapedTypeComponents(outShape, resultTy.getElementType()));
-  return ret;
+  return LogicalResult::success(ret);
```
LongshengDu, line 36: What is the difference between `return ret` and `return LogicalResult::success(ret)`

lib/gc/Dialect/OneDNNGraph/OneDNNGraphOps.cpp
```diff
@@ -65,11 +27,13 @@ LogicalResult onednn_graph::AddOp::inferReturnTypeComponents(
   auto getShapeIdx = [](ValueShapeRange operands, size_t i) {
     return operands.getShape(i);
   };
-  auto ret =
-      inferBroadcastShape<ValueShapeRange>(operands, outShape, getShapeIdx);
+  llvm::SmallVector<int64_t> input1, input2;
+  getShapeIdx(operands, 0).getDims(input1);
```
LongshengDu, line 31: Can we use something like `getShape` to get the ArrayRef of shape for performance?

lib/gc/Dialect/OneDNNGraph/OneDNNGraphOps.cpp
```diff
@@ -158,22 +122,22 @@ LogicalResult onednn_graph::MatMulOp::inferReturnTypeComponents(
...
-  };
   // final shape
   auto retShape = ShapedTypeComponents(outShape, lhsShape.getElementType());
   inferredReturnShapes.push_back(retShape);
   // check for bias broadcasting
   if (adaptor.getBias()) {
     ShapeAdaptor biasShape(adaptor.getBias().getType());
     ShapeAdaptor matShape(retShape);
+    llvm::SmallVector<int64_t> matmulShape;
+    matShape.getDims(matmulShape);
```
LongshengDu, line 133: Can we use something like `getShape` to get the ArrayRef of shape for performance?

lib/gc/Dialect/OneDNNGraph/OneDNNGraphOps.cpp
BRUCE11111: The return value ret returned by the new function is `bool` . Here we need to convert it to `logicalResult`.

LongshengDu: Oh, I see, to convert bool to LogicalResult

lib/gc/Dialect/OneDNNGraph/OneDNNGraphOps.cpp
BRUCE11111: Thanks~ Optimized the code~

LongshengDu: Also, maybe `dyn_cast<ShapedType>()` should be better, in case we need to support other shaped types in the future.

LongshengDu (approved):

include/gc/Dialect/OneDNNGraph/OneDNNGraphTypes.td
```diff
@@ -17,14 +17,26 @@ include "OneDNNGraphDialect.td"
...
+def OneDNNGraph_Float : AnyTypeOf<[F32,
+			                       F16,
+			                       BF16]>;
+
+//===----------------------------------------------------------------------===//
+// Integer types.
+//===----------------------------------------------------------------------===//
+
+def OneDNNGraph_Int : AnyTypeOf<[SI<8>,
+                                 UI<8>]>;
```
kurapov-peter, line 32: What is the rationale for having separate types? Also, why do we only have fp16 and int8?

kurapov-peter (changes requested):

include/gc/Dialect/OneDNNGraph/OneDNNGraphTypes.td
LongshengDu: Easier to construct Quantize/Dequantize ops later. https://oneapi-src.github.io/oneDNN/dev_guide_op_quantize.html#supported-data-types

ZhennanQin (approved):

kurapov-peter (approved):
Summary: having separate types just follows the spec and helps us automatically check typing. The change associated with potential op semantic change (e.g., quantize int->int) is estimated as minor.

kurapov-peter:
This will still need to change when fp8eX types are added.

Issue #54, opened 2024-05-09
Support fine-grain fusion


ZhennanQin:
- Matmul fusion with element-wise/reduce/broadcast ops.
- Pre-op and post-op fusion.
- Multi-consumer and multi-producer support.
- Cost-model to determine whether to fuse or not.

PR #84, opened 2024-05-16
[WIP] Tiling

Devjiu:


ZhennanQin:
Would you please provide more background about this work?

Devjiu:
> Would you please provide more background about this work?

As far as I know we don't have aligned approach to tiling. We have an approach that was used in v1 with fusion anchors and separation of coarse-grained and fine-grained fusion.  

Also it's not clear to me how tiling approaches should be united with packing approaches that required for deep nested matmul optimization.  

So in this PR I am investigating possibilities to have something similar with TileLayout to unify coarse and fine grain and to estimate will it provide similar resulting IR. 

It's proof of concept for TileLayout approach to discuss it in more detail. (I think we need align, before implementing #54)

This PR still not ready for review.

ZhennanQin:
> > Would you please provide more background about this work?
> 
> As far as I know we don't have aligned approach to tiling. We have an approach that was used in v1 with fusion anchors and separation of coarse-grained and fine-grained fusion.
> 
> Also it's not clear to me how tiling approaches should be united with packing approaches that required for deep nested matmul optimization.
> 
> So in this PR I am investigating possibilities to have something similar with TileLayout to unify coarse and fine grain and to estimate will it provide similar resulting IR.
> 
> It's proof of concept for TileLayout approach to discuss it in more detail. (I think we need align, before implementing #54)
> 
> This PR still not ready for review.

RFC is the suggested way to align the design, please discuss your concerns in existing RFC, or prepare new RFC if you have a completely new design. We haven't finished the coarse-grain fusion design yet, and we will finish it after we have fine-grain fusion implementation which can provide more insights.

PR #85, opened 2024-05-17
action: add license check

WangJialei-A:
Add license check
Files in the changeset with suffix `.c`, `.cpp`, `.h`, `.hpp`, `.py`, `CMakeLists.txt` will be checked.
Please see the result at #86 and [url](https://github.com/intel/graph-compiler/actions/runs/9124527004/job/25088869295?pr=86)

### DIFF
--- a/.github/workflows/license.yml
+++ b/.github/workflows/license.yml
@@ -1,0 +1,32 @@
+name: Graph Compiler License Check
+
+on:
+  pull_request:
+
+permissions: read-all
+
+jobs:
+  license-check:
+    name: License Check
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+      repository-projects: write
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          submodules: true
+
+      - name: Get merge base
+        run: |
+          echo "MERGE_BASE=`git merge-base ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }}`" >> $GITHUB_ENV
+
+      - name: Get changed files
+        run: |
+          echo "CHANGED_FILES=`git diff --name-only $MERGE_BASE ${{ github.event.pull_request.head.sha }} | paste -sd,`" >> $GITHUB_ENV
+
+      - name: Perform license check
+        run: "python scripts/license.py --files $CHANGED_FILES"
--- a/scripts/license.py
+++ b/scripts/license.py
@@ -1,0 +1,164 @@
+# Copyright (C) 2024 Intel Corporation
+# 
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# 
+# http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+# 
+# SPDX-License-Identifier: Apache-2.0
+
+import datetime, sys, re, argparse
+from typing import Dict, Set
+
+WIDTH: int = 80
+intel_license: list[str] = [
+    'Copyright \\(C\\) (\\d\\d\\d\\d-)?$YEAR Intel Corporation',
+    '',
+    'Licensed under the Apache License, Version 2.0 (the "License");',
+    'you may not use this file except in compliance with the License.',
+    'You may obtain a copy of the License at',
+    '',
+    'http://www.apache.org/licenses/LICENSE-2.0',
+    '',
+    'Unless required by applicable law or agreed to in writing,',
+    'software distributed under the License is distributed on an "AS IS" BASIS,',
+    'WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.',
+    'See the License for the specific language governing permissions',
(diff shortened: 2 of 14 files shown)


.github/workflows/license.yml
```diff
@@ -0,0 +1,81 @@
...
+      - name: Get changed files
+        run: |
+          echo "CHANGED_FILES=`git diff --name-only $MERGE_BASE ${{ github.event.pull_request.head.sha }} | paste -sd,`" >> $GITHUB_ENV
+
+      - name: Perform license check
+        shell: python
+        run: |
+            import os, datetime, sys
+
+            intel_license: list[str] = [
```
Menooker, line 36: shall we put this script in `.github` or somewhere inside our repo?

.github/workflows/license.yml
```diff
@@ -0,0 +1,81 @@
...
+                'Unless required by applicable law or agreed to in writing,',
+                'software distributed under the License is distributed on an "AS IS" BASIS,',
+                'WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.',
+                'See the License for the specific language governing permissions',
+                'and limitations under the License.',
+                'SPDX-License-Identifier: Apache-2.0',
+            ]
+
+            llvm_license: list[str] = [
+                'This file is licensed under the Apache License v2.0 with LLVM Exceptions.',
```
Menooker, line 51: LLVM's header looks like

```
//===- Example.cpp - Tests for Example ------------------------------------===//
//
// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//
```

Shall we check the header `//===- Example.cpp - Tests for Example ------------------------------------===//` and the footer `//===----------------------------------------------------------------------===//` as well?

Note that the header has variable number of `------------------------------------` to align the line wrap limit.

.github/workflows/license.yml
WangJialei-A: @Menooker 
If developers need to run it as a local check, it is ok to split the script into a seperate python file. 
Do you really need to do the local check?

.github/workflows/license.yml
```diff
@@ -0,0 +1,81 @@
...
+                'See https://llvm.org/LICENSE.txt for license information.',
+                'SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception',
+            ]
+
+            def check_license(filename: str, license: list[str]):
```
Menooker, line 56: oneDNN's script also provides a utility to auto fix the headers. It is useful when a PR is across two years (to update year). 
I think for LLVM, auto fixing headers are also useful, when we introduce many cpp files in a PR, and we need to manually fill-in the file names in the licence. Maybe we can improve the script in another PR.

.github/workflows/license.yml
WangJialei-A: @Menooker 
The string `Tests for Example` you mentioned is not a costant string or from the filename. I do not know how to verify it.
That's why I skipped the first line check for llvm license.

.github/workflows/license.yml
Menooker: Please also note that we already have `script/` directory for compiling the repo and compiling LLVM. The `sh` files are used in CI too. I think we should add the py file to this directory.

.github/workflows/license.yml
WangJialei-A: We can write another small script to handle the fixing job in the future, but not in the automatic CI script.

.github/workflows/license.yml
Menooker: It is ` {file name} - {desc} ---------`. I think we can check the filename and the number of `---------`. For auto-fix, we can fill-in the filename and a place holder "{DESC}" and auto fix the number of `------------`.

In checking mode, we can throw an error when detecting "{DESC}".

A workflow for the user to use auto-fix should be:

1. write code without LLVM header
2. auto-fix for the first time
3. fill in {DESC} in the code
4. auto-fix again, maybe check if there are remaining {DESC}

.github/workflows/license.yml
AshburnLee: How about put this check script into `repo_dir/scripts/license_check.py`, so that this script can be used not only in actions but also from local.

test/lit.cfg.py
```diff
@@ -1,3 +1,19 @@
+# Copyright (C) 2024 Intel Corporation
+#
```
Menooker, line 2: It is part of MLIR testing. And upstream MLIR does not add copyright on this file

scripts/license.py
```diff
@@ -0,0 +1,110 @@
...
+                var['$LANG'] = "" 
+                break
+    if "$LANG" not in var:
+        continue
+
+    if not check_license(filepath, lic, var, re_line):
+        success = False
+        print("Fail   : %s" % filepath)
+    else:
+        print("Success: %s" % filepath)
```
Menooker, line 108: Can we provide suggestions on how to fix that?

test/lit.cfg.py
WangJialei-A: > It is part of MLIR testing. And upstream MLIR does not add copyright on this file

@Menooker  It is Intel license, not the LLVM one and non-related with upstream work.

Menooker: It is copied and modified based on upstream MLIR. In upstream, this file seems not to have copyright?

WangJialei-A:
@Menooker  
Please start another review. You can see the test result here: [url](https://github.com/intel/graph-compiler/actions/runs/9169282144/job/25209456564)

Menooker (approved):

Issue #14, opened 2024-04-30
Add linalgx dialect definition


kurapov-peter:
This one is for the `matmul_nd`, potentially lowering back to linalg.

PR #88, opened 2024-05-20
[Dialect] Add basic linalgx dialect infrastructure and linalgx.sigmoid op

LongshengDu:
Add basic linalgx dialect infrastructure.  

Kept a copy of mlir `LinalgOps.cpp` helpers inside our project. 

Added linalgx.sigmoid as a structured named op, so now linalgx.sigmoid can be lowered to one single linalg.generic using `--linalg-generalize-named-ops`.

A potential `linalg-decomposition` pass can be a idea to decomposed linalgx.sigmoid to a series of named ops.

Tracking: #14


lib/gc/Dialect/Linalgx/LinalgxOps.cpp
```diff
@@ -10,5 +10,82 @@
 #include "gc/Dialect/Linalgx/LinalgxDialect.h"
 #include "mlir/IR/OpImplementation.h"
 
+//===----------------------------------------------------------------------===//
+// Builder helper from mlir/lib/Dialect/Linalg/IR/LinalgOps.cpp
+//===----------------------------------------------------------------------===//
+
+#include "LinalgOps.cpp.inc"
```
ciyongch, line 17: Can we include the original file from MLIR and not keeping a partial copy as it might diverge as the main stream moving forward.

LongshengDu: The file cannot be included after MLIR is built and the helper functions are static and cannot used in other files, so a partial copy is kept.

kurapov-peter (commented):
Since we are going to lower sigmoid into a set of math+arith do we need it at all in linalgx? Can we just generate the generic right from ondenn graph?

LongshengDu:
> Since we are going to lower sigmoid into a set of math+arith do we need it at all in linalgx? Can we just generate the generic right from ondenn graph?

We want to experiment with all named ops first and learn the strength/limitations, in the future we will decide how to continue the design for future/upstream, etc.

include/gc/Dialect/Linalgx/LinalgxDialect.td
```diff
@@ -24,7 +24,27 @@ def LinalgxDialect : Dialect {
     }];
     let cppNamespace = "::mlir::linalgx";
 
-    let useDefaultTypePrinterParser = 1;
+    let dependentDialects = [
+        "arith::ArithDialect",
+        "affine::AffineDialect",
```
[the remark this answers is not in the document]
LongshengDu: This is a copy of the linalg define, just to keep the same.

lib/gc/Dialect/Linalgx/LinalgOps.cpp.inc
```diff
@@ -0,0 +1,544 @@
+//===- LinalgOps.cpp - Implementation of the linalg operations ------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file is a partial copy of mlir/lib/Dialect/Linalg/IR/LinalgOps.cpp
```
kurapov-peter, line 9: Is it possible to directly use the original instead?

.github/workflows/clang-tidy.yml
kurapov-peter, line 1: Seems unrelated? Could we push it separately?

kurapov-peter (approved):
If possible, let's get rid of the duplication + some minor comments. Otherwise fine.

.github/workflows/clang-tidy.yml
LongshengDu: This PR generate multiple inc target in one makefile, so this is changed to accommodate it.

lib/gc/Dialect/Linalgx/LinalgOps.cpp.inc
LongshengDu: Has not found another way. The file cannot be included after MLIR is built and the class/helper functions are static and cannot be used in other files.

Issue #53, opened 2024-05-09
nested matmul implementation


ZhennanQin:
- Support FP32 and BF16.
- Block layout first, then extend to plain layout.
- Default config to cover shapes used in LLAMA2.

PR #90, opened 2024-05-20
[Transform][Tiling] Add deep tile support for matmul

zhczhong:
Tracking https://github.com/intel/graph-compiler/issues/53

**TODO**:
- [X] the nested outer loop generation
- [X] partial reduction support
- - [X] enhance the PartialReductionOpInterface to allow user control where the new parallel dims are inserted
- - [X] Erase the reducant linalg.FillOp in partial reduction
- [X] merge all parallel iterator into a single scf.forall before nested parallel is ready
- [X] fuse the linalg.fillOp into the innermost loop body
- [X] replace all genericOp with linalg named op
- [X] Support 4Dx4/5D->4D, 2Dx2D->2D, 2Dx4/5D->2D
- [X] Dtype Support(f32, bf16)
- [X] Fuse the f32->bf16 cast into the last loop about K axis
- [ ] Support Batch matmul 
- [ ] Balance211 support
- [ ] Tune a general matmul config based on cost model
- [x] Fuse the linalg.copy to the innermost loop


zhczhong:
Support use linalgx.batch_reduce_vnni(bf16xbf16->f32) and fuse the cast(f32->bf16) to the last loop about K axis
```mlir
func.func @matmul_4Dx4D_bf16(%arg0: tensor<128x128x32x32xbf16>, %arg1: tensor<128x128x16x32x2xbf16>) -> tensor<128x128x32x32xbf16> {
    %cst_0 = arith.constant 0.000000e+00 : bf16
    %0 = tensor.empty() : tensor<128x128x32x32xbf16>
    %1 = linalg.fill ins(%cst_0 : bf16) outs(%0 : tensor<128x128x32x32xbf16>) -> tensor<128x128x32x32xbf16>
    %2 = linalgx.mm4d_vnni ins(%arg0, %arg1 : tensor<128x128x32x32xbf16>, tensor<128x128x16x32x2xbf16>) outs(%1 : tensor<128x128x32x32xbf16>)  -> tensor<128x128x32x32xbf16>
    return %2 : tensor<128x128x32x32xbf16>
}
```
will be transformed into 
```mlir
#map = affine_map<(d0) -> (d0 * 64)>
#map1 = affine_map<(d0)[s0, s1] -> (d0 * 64 + s0 + s1)>
module {
  func.func @matmul_4Dx4D_bf16(%arg0: tensor<128x128x32x32xbf16>, %arg1: tensor<128x128x16x32x2xbf16>) -> tensor<128x128x32x32xbf16> {
    %c1 = arith.constant 1 : index
    %c128 = arith.constant 128 : index
    %c2 = arith.constant 2 : index
    %c64 = arith.constant 64 : index
    %c0 = arith.constant 0 : index
    %cst = arith.constant 0.000000e+00 : bf16
    %0 = tensor.empty() : tensor<128x128x32x32xbf16>
    %1 = scf.forall (%arg2, %arg3) in (2, 2) shared_outs(%arg4 = %0) -> (tensor<128x128x32x32xbf16>) {
      %2 = affine.apply #map(%arg2)
      %3 = affine.apply #map(%arg3)
      %extracted_slice = tensor.extract_slice %arg4[%2, %3, 0, 0] [64, 64, 32, 32] [1, 1, 1, 1] : tensor<128x128x32x32xbf16> to tensor<64x64x32x32xbf16>
      %4 = scf.for %arg5 = %c0 to %c64 step %c2 iter_args(%arg6 = %extracted_slice) -> (tensor<64x64x32x32xbf16>) {
        %extracted_slice_0 = tensor.extract_slice %arg6[%arg5, 0, 0, 0] [2, 64, 32, 32] [1, 1, 1, 1] : tensor<64x64x32x32xbf16> to tensor<2x64x32x32xbf16>
        %7 = scf.for %arg7 = %c0 to %c64 step %c2 iter_args(%arg8 = %extracted_slice_0) -> (tensor<2x64x32x32xbf16>) {
          %extracted_slice_1 = tensor.extract_slice %arg8[0, %arg7, 0, 0] [2, 2, 32, 32] [1, 1, 1, 1] : tensor<2x64x32x32xbf16> to tensor<2x2x32x32xbf16>
          %8 = tensor.empty() : tensor<2x2x32x32xf32>
          %9 = scf.for %arg9 = %c0 to %c128 step %c2 iter_args(%arg10 = %8) -> (tensor<2x2x32x32xf32>) {
            %11 = scf.for %arg11 = %c0 to %c2 step %c1 iter_args(%arg12 = %arg10) -> (tensor<2x2x32x32xf32>) {
              %extracted_slice_3 = tensor.extract_slice %arg12[%arg11, 0, 0, 0] [1, 2, 32, 32] [1, 1, 1, 1] : tensor<2x2x32x32xf32> to tensor<1x2x32x32xf32>
              %12 = scf.for %arg13 = %c0 to %c2 step %c1 iter_args(%arg14 = %extracted_slice_3) -> (tensor<1x2x32x32xf32>) {
                %13 = affine.apply #map1(%arg2)[%arg11, %arg5]
                %extracted_slice_5 = tensor.extract_slice %arg0[%13, %arg9, 0, 0] [1, 2, 32, 32] [1, 1, 1, 1] : tensor<128x128x32x32xbf16> to tensor<2x32x32xbf16>
                %14 = affine.apply #map1(%arg3)[%arg13, %arg7]
                %extracted_slice_6 = tensor.extract_slice %arg1[%14, %arg9, 0, 0, 0] [1, 2, 16, 32, 2] [1, 1, 1, 1, 1] : tensor<128x128x16x32x2xbf16> to tensor<2x16x32x2xbf16>
                %extracted_slice_7 = tensor.extract_slice %arg14[0, %arg13, 0, 0] [1, 1, 32, 32] [1, 1, 1, 1] : tensor<1x2x32x32xf32> to tensor<32x32xf32>
                %15 = arith.cmpi eq, %arg9, %c0 : index
                %16 = scf.if %15 -> (tensor<32x32xf32>) {
                  %17 = linalg.fill ins(%cst : bf16) outs(%extracted_slice_7 : tensor<32x32xf32>) -> tensor<32x32xf32>
                  %18 = linalgx.batch_reduce_matmul_vnni ins(%extracted_slice_5, %extracted_slice_6 : tensor<2x32x32xbf16>, tensor<2x16x32x2xbf16>) outs(%17 : tensor<32x32xf32>) -> tensor<32x32xf32>
                  scf.yield %18 : tensor<32x32xf32>
                } else {
                  %17 = linalgx.batch_reduce_matmul_vnni ins(%extracted_slice_5, %extracted_slice_6 : tensor<2x32x32xbf16>, tensor<2x16x32x2xbf16>) outs(%extracted_slice_7 : tensor<32x32xf32>) -> tensor<32x32xf32>
                  scf.yield %17 : tensor<32x32xf32>
                }
                %inserted_slice_8 = tensor.insert_slice %16 into %arg14[0, %arg13, 0, 0] [1, 1, 32, 32] [1, 1, 1, 1] : tensor<32x32xf32> into tensor<1x2x32x32xf32>
                scf.yield %inserted_slice_8 : tensor<1x2x32x32xf32>
              }
              %inserted_slice_4 = tensor.insert_slice %12 into %arg12[%arg11, 0, 0, 0] [1, 2, 32, 32] [1, 1, 1, 1] : tensor<1x2x32x32xf32> into tensor<2x2x32x32xf32>
              scf.yield %inserted_slice_4 : tensor<2x2x32x32xf32>
            }
            scf.yield %11 : tensor<2x2x32x32xf32>
          }
          %10 = linalg.copy ins(%9 : tensor<2x2x32x32xf32>) outs(%extracted_slice_1 : tensor<2x2x32x32xbf16>) -> tensor<2x2x32x32xbf16>
          %inserted_slice_2 = tensor.insert_slice %10 into %arg8[0, %arg7, 0, 0] [2, 2, 32, 32] [1, 1, 1, 1] : tensor<2x2x32x32xbf16> into tensor<2x64x32x32xbf16>
          scf.yield %inserted_slice_2 : tensor<2x64x32x32xbf16>
        }
        %inserted_slice = tensor.insert_slice %7 into %arg6[%arg5, 0, 0, 0] [2, 64, 32, 32] [1, 1, 1, 1] : tensor<2x64x32x32xbf16> into tensor<64x64x32x32xbf16>
        scf.yield %inserted_slice : tensor<64x64x32x32xbf16>
      }
      %5 = affine.apply #map(%arg2)
      %6 = affine.apply #map(%arg3)
      scf.forall.in_parallel {
        tensor.parallel_insert_slice %4 into %arg4[%5, %6, 0, 0] [64, 64, 32, 32] [1, 1, 1, 1] : tensor<64x64x32x32xbf16> into tensor<128x128x32x32xbf16>
      }
    }
    return %1 : tensor<128x128x32x32xbf16>
  }
}
```

lib/gc/Transforms/DeepTileContractionNamedOp.cpp
```diff
@@ -0,0 +1,973 @@
...
+  SmallVector<unsigned> idxList;
+  for (auto [idx, type] : llvm::enumerate(tyList)) {
+    if (type == ty) {
+      idxList.push_back(idx);
+    }
+  }
+  return idxList;
+}
+
+MatmulConfig getDefaultMatmulConfig(linalg::LinalgOp &linalgOp) {
```
yifeizh2, line 104: Can you expose this method in the future, as it is related to layout inference in global layout analysis pass.

lib/gc/Transforms/DeepTileContractionNamedOp.cpp
```diff
@@ -0,0 +1,973 @@
...
+#include <memory>
+
+namespace mlir {
+namespace gc {
+#define GEN_PASS_DEF_DEEPTILECONTRACTIONNAMEDOP
+#include "gc/Transforms/Passes.h.inc"
+
+namespace {
+
+struct SystemDesc {
```
yifeizh2, line 48: Will this struct be replaced with target description in future?

lib/gc/Transforms/DeepTileContractionNamedOp.cpp
zhczhong: ok, I will expose this method later

zhczhong:
**Update**: Fuse the cast(f32->bf16) to the innermost loop
```mlir
func.func @matmul_4Dx4D_bf16(%arg0: tensor<128x128x32x32xbf16>, %arg1: tensor<128x128x16x32x2xbf16>) -> tensor<128x128x32x32xbf16> {
    %cst_0 = arith.constant 0.000000e+00 : bf16
    %0 = tensor.empty() : tensor<128x128x32x32xbf16>
    %1 = linalg.fill ins(%cst_0 : bf16) outs(%0 : tensor<128x128x32x32xbf16>) -> tensor<128x128x32x32xbf16>
    %2 = linalgx.mm4d_vnni ins(%arg0, %arg1 : tensor<128x128x32x32xbf16>, tensor<128x128x16x32x2xbf16>) outs(%1 : tensor<128x128x32x32xbf16>)  -> tensor<128x128x32x32xbf16>
    return %2 : tensor<128x128x32x32xbf16>
}
```
will be transformed to
```mlir
#map = affine_map<(d0) -> (d0 * 64)>
#map1 = affine_map<(d0)[s0, s1] -> (d0 * 64 + s0 + s1)>
module {
  func.func @matmul_4Dx4D_bf16(%arg0: tensor<128x128x32x32xbf16>, %arg1: tensor<128x128x16x32x2xbf16>) -> tensor<128x128x32x32xbf16> {
    %c1 = arith.constant 1 : index
    %c128 = arith.constant 128 : index
    %c2 = arith.constant 2 : index
    %c64 = arith.constant 64 : index
    %c0 = arith.constant 0 : index
    %cst = arith.constant 0.000000e+00 : bf16
    %0 = tensor.empty() : tensor<128x128x32x32xbf16>
    %1 = scf.forall (%arg2, %arg3) in (2, 2) shared_outs(%arg4 = %0) -> (tensor<128x128x32x32xbf16>) {
      %2 = affine.apply #map(%arg2)
      %3 = affine.apply #map(%arg3)
      %extracted_slice = tensor.extract_slice %arg4[%2, %3, 0, 0] [64, 64, 32, 32] [1, 1, 1, 1] : tensor<128x128x32x32xbf16> to tensor<64x64x32x32xbf16>
      %4 = scf.for %arg5 = %c0 to %c64 step %c2 iter_args(%arg6 = %extracted_slice) -> (tensor<64x64x32x32xbf16>) {
        %extracted_slice_0 = tensor.extract_slice %arg6[%arg5, 0, 0, 0] [2, 64, 32, 32] [1, 1, 1, 1] : tensor<64x64x32x32xbf16> to tensor<2x64x32x32xbf16>
        %7 = scf.for %arg7 = %c0 to %c64 step %c2 iter_args(%arg8 = %extracted_slice_0) -> (tensor<2x64x32x32xbf16>) {
          %extracted_slice_1 = tensor.extract_slice %arg8[0, %arg7, 0, 0] [2, 2, 32, 32] [1, 1, 1, 1] : tensor<2x64x32x32xbf16> to tensor<2x2x32x32xbf16>
          %8 = tensor.empty() : tensor<2x2x32x32xf32>
          %9:2 = scf.for %arg9 = %c0 to %c128 step %c2 iter_args(%arg10 = %8, %arg11 = %extracted_slice_1) -> (tensor<2x2x32x32xf32>, tensor<2x2x32x32xbf16>) {
            %10:2 = scf.for %arg12 = %c0 to %c2 step %c1 iter_args(%arg13 = %arg10, %arg14 = %arg11) -> (tensor<2x2x32x32xf32>, tensor<2x2x32x32xbf16>) {
              %extracted_slice_3 = tensor.extract_slice %arg13[%arg12, 0, 0, 0] [1, 2, 32, 32] [1, 1, 1, 1] : tensor<2x2x32x32xf32> to tensor<1x2x32x32xf32>
              %extracted_slice_4 = tensor.extract_slice %arg14[%arg12, 0, 0, 0] [1, 2, 32, 32] [1, 1, 1, 1] : tensor<2x2x32x32xbf16> to tensor<1x2x32x32xbf16>
              %11:2 = scf.for %arg15 = %c0 to %c2 step %c1 iter_args(%arg16 = %extracted_slice_3, %arg17 = %extracted_slice_4) -> (tensor<1x2x32x32xf32>, tensor<1x2x32x32xbf16>) {
                %12 = affine.apply #map1(%arg2)[%arg12, %arg5]
                %extracted_slice_7 = tensor.extract_slice %arg0[%12, %arg9, 0, 0] [1, 2, 32, 32] [1, 1, 1, 1] : tensor<128x128x32x32xbf16> to tensor<2x32x32xbf16>
                %13 = affine.apply #map1(%arg3)[%arg15, %arg7]
                %extracted_slice_8 = tensor.extract_slice %arg1[%13, %arg9, 0, 0, 0] [1, 2, 16, 32, 2] [1, 1, 1, 1, 1] : tensor<128x128x16x32x2xbf16> to tensor<2x16x32x2xbf16>
                %extracted_slice_9 = tensor.extract_slice %arg16[0, %arg15, 0, 0] [1, 1, 32, 32] [1, 1, 1, 1] : tensor<1x2x32x32xf32> to tensor<32x32xf32>
                %extracted_slice_10 = tensor.extract_slice %arg17[0, %arg15, 0, 0] [1, 1, 32, 32] [1, 1, 1, 1] : tensor<1x2x32x32xbf16> to tensor<32x32xbf16>
                %14 = arith.cmpi eq, %arg9, %c0 : index
                %15 = scf.if %14 -> (tensor<32x32xf32>) {
                  %18 = linalg.fill ins(%cst : bf16) outs(%extracted_slice_9 : tensor<32x32xf32>) -> tensor<32x32xf32>
                  %19 = linalgx.batch_reduce_matmul_vnni ins(%extracted_slice_7, %extracted_slice_8 : tensor<2x32x32xbf16>, tensor<2x16x32x2xbf16>) outs(%18 : tensor<32x32xf32>) -> tensor<32x32xf32>
                  scf.yield %19 : tensor<32x32xf32>
                } else {
                  %18 = linalgx.batch_reduce_matmul_vnni ins(%extracted_slice_7, %extracted_slice_8 : tensor<2x32x32xbf16>, tensor<2x16x32x2xbf16>) outs(%extracted_slice_9 : tensor<32x32xf32>) -> tensor<32x32xf32>
                  scf.yield %18 : tensor<32x32xf32>
                }
                %16 = arith.cmpi eq, %arg9, %c0 : index
                %17 = scf.if %16 -> (tensor<32x32xbf16>) {
                  %18 = linalg.copy ins(%15 : tensor<32x32xf32>) outs(%extracted_slice_10 : tensor<32x32xbf16>) -> tensor<32x32xbf16>
                  scf.yield %18 : tensor<32x32xbf16>
                } else {
                  scf.yield %extracted_slice_10 : tensor<32x32xbf16>
                }
                %inserted_slice_11 = tensor.insert_slice %15 into %arg16[0, %arg15, 0, 0] [1, 1, 32, 32] [1, 1, 1, 1] : tensor<32x32xf32> into tensor<1x2x32x32xf32>
                %inserted_slice_12 = tensor.insert_slice %17 into %arg17[0, %arg15, 0, 0] [1, 1, 32, 32] [1, 1, 1, 1] : tensor<32x32xbf16> into tensor<1x2x32x32xbf16>
                scf.yield %inserted_slice_11, %inserted_slice_12 : tensor<1x2x32x32xf32>, tensor<1x2x32x32xbf16>
              }
              %inserted_slice_5 = tensor.insert_slice %11#0 into %arg13[%arg12, 0, 0, 0] [1, 2, 32, 32] [1, 1, 1, 1] : tensor<1x2x32x32xf32> into tensor<2x2x32x32xf32>
              %inserted_slice_6 = tensor.insert_slice %11#1 into %arg14[%arg12, 0, 0, 0] [1, 2, 32, 32] [1, 1, 1, 1] : tensor<1x2x32x32xbf16> into tensor<2x2x32x32xbf16>
              scf.yield %inserted_slice_5, %inserted_slice_6 : tensor<2x2x32x32xf32>, tensor<2x2x32x32xbf16>
            }
            scf.yield %10#0, %10#1 : tensor<2x2x32x32xf32>, tensor<2x2x32x32xbf16>
          }
          %inserted_slice_2 = tensor.insert_slice %9#1 into %arg8[0, %arg7, 0, 0] [2, 2, 32, 32] [1, 1, 1, 1] : tensor<2x2x32x32xbf16> into tensor<2x64x32x32xbf16>
          scf.yield %inserted_slice_2 : tensor<2x64x32x32xbf16>
        }
        %inserted_slice = tensor.insert_slice %7 into %arg6[%arg5, 0, 0, 0] [2, 64, 32, 32] [1, 1, 1, 1] : tensor<2x64x32x32xbf16> into tensor<64x64x32x32xbf16>
        scf.yield %inserted_slice : tensor<64x64x32x32xbf16>
      }
      %5 = affine.apply #map(%arg2)
      %6 = affine.apply #map(%arg3)
      scf.forall.in_parallel {
        tensor.parallel_insert_slice %4 into %arg4[%5, %6, 0, 0] [64, 64, 32, 32] [1, 1, 1, 1] : tensor<64x64x32x32xbf16> into tensor<128x128x32x32xbf16>
      }
    }
    return %1 : tensor<128x128x32x32xbf16>
  }
}
```

include/gc/IR/EasyBuildSCF.h
```diff
@@ -0,0 +1,187 @@
...
+};
+inline int IfIterator::operator*() const {
+  auto &blocks = ptr->op->getRegion(index);
+  ptr->s->builder.setInsertionPointToStart(&blocks.back());
+  return index;
+}
+
+} // namespace impl
+
+impl::IfSimulator makeIfRange(const EasyBuilder &s, Operation *op) {
```
yifeizh2, line 153: Missing an `inline` here

include/gc/Dialect/Arith/Utils/EasyBuild.h
```diff
@@ -0,0 +1,432 @@
...
+
+#undef DEF_EASYBUILD_BINARY_OPERATOR_FOR_INT
+#undef DEF_EASYBUILD_BINARY_OPERATOR_FOR_TYPE
+
+inline EBFloatPoint operator-(const EBFloatPoint &a) {
+  return OperatorHandlers::create<EBFloatPoint, arith::NegFOp>(a.builder, a.v);
+}
+
+#define DEF_EASYBUILD_CMP_OPERATOR(OP, OPCLASS, TYPE, PRED)                    \
+  EBUnsigned operator OP(const TYPE &a, const TYPE &b) {                       \
```
yifeizh2, line 267: Missing an `inline` here

lib/gc/Transforms/CMakeLists.txt
yifeizh2, line 1: @xurui1995 found the following undefined reference issue while building the repo
```
../../../../lib/libGCPasses.a(DeepTileContractionNamedOp.cpp.o): In function `mlir::gc::(anonymous namespace)::deepTileMatmul::innerBodyGeneration(mlir::RewriterBase&, mlir::linalg::LinalgOp, mlir::linalg::LinalgOp, mlir::gc::(anonymous namespace)::deepTileMatmul::innerBodyGenerationOption&) const':
DeepTileContractionNamedOp.cpp:(.text._ZNK4mlir2gc12_GLOBAL__N_114deepTileMatmul19innerBodyGenerationERNS_12RewriterBaseENS_6linalg8LinalgOpES6_RNS2_25innerBodyGenerationOptionE+0xab): undefined reference to `mlir::gc::MatmulConfigAnalysis::MatmulConfigAnalysis(mlir::Operation*)'
../../../../lib/libGCPasses.a(DeepTileContractionNamedOp.cpp.o): In function `mlir::gc::(anonymous namespace)::deepTileMatmul::matchAndRewrite(mlir::linalg::LinalgOp, mlir::PatternRewriter&) const':
DeepTileContractionNamedOp.cpp:(.text._ZNK4mlir2gc12_GLOBAL__N_114deepTileMatmul15matchAndRewriteENS_6linalg8LinalgOpERNS_15PatternRewriterE+0x1e8): undefined reference to `mlir::gc::MatmulConfigAnalysis::MatmulConfigAnalysis(mlir::Operation*)'
collect2: error: ld returned 1 exit status
gmake[3]: *** [test/mlir/unittests/ExecutionEngine/CMakeFiles/GCExecutionEngineTests.dir/build.make:361: test/mlir/unittests/ExecutionEngine/GCExecutionEngineTests] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:6761: test/mlir/unittests/ExecutionEngine/CMakeFiles/GCExecutionEngineTests.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:6421: test/mlir/test/CMakeFiles/gc-check.dir/rule] Error 2
```

The solution is to add `GCAnalysis` to `LINK_LIBS`. Please help update. Thanks!!!

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,443 @@
...
+  });
+  auto thresholdCost = costs[idx[(size_t)(eliminationRatio * configs.size())]];
+  thresholdCost =
+      threshold < thresholdCost && threshold > 0 ? threshold : thresholdCost;
+  for (size_t i = 0; i < configs.size(); i++) {
+    if (costs[idx[i]] <= thresholdCost) {
+      result.push_back(configs[idx[i]]);
+    }
+  }
+  llvm::errs() << "thresholdCost is: " << thresholdCost
```
yifeizh2, line 194: Will you change it to  `LLVM_DEBUG(llvm::dbgs() xxxx)`?

zhczhong: Thanks for reminding! I will warp these with `LLVM_DEBUG`

ZhennanQin:
Please rebase and polish it for code-review.

include/gc/Analysis/MatmulConfigAnalysis.h
```diff
@@ -0,0 +1,123 @@
...
+  for (auto [idx, type] : llvm::enumerate(tyList)) {
+    if (type == ty) {
+      idxList.push_back(idx);
+    }
+  }
+  return idxList;
+}
+
+static FailureOr<SmallVector<SmallVector<DimType>>>
+getOprandDimType(linalg::LinalgOp &linalgOp) {
```
yifeizh2, line 84: Please consider also extend to *Transpose*Op series such as the following
```
// ...
  } else if (llvm::isa<linalg::MatmulTransposeAOp>(linalgOp)) {
    return SmallVector<SmallVector<DimType>>{
        SmallVector<DimType>{DimType::K, DimType::M},
        SmallVector<DimType>{DimType::K, DimType::N},
        SmallVector<DimType>{DimType::M, DimType::N}};
  } else if (llvm::isa<linalg::MatmulTransposeBOp>(linalgOp)) {
    return SmallVector<SmallVector<DimType>>{
        SmallVector<DimType>{DimType::M, DimType::K},
        SmallVector<DimType>{DimType::N, DimType::K},
        SmallVector<DimType>{DimType::M, DimType::N}};
  } else if (llvm::isa<linalg::BatchMatmulTransposeAOp>(linalgOp)) {
    return SmallVector<SmallVector<DimType>>{
        SmallVector<DimType>{DimType::Batch, DimType::K, DimType::M},
        SmallVector<DimType>{DimType::Batch, DimType::K, DimType::N},
        SmallVector<DimType>{DimType::Batch, DimType::M, DimType::N}};
  } else if (llvm::isa<linalg::BatchMatmulTransposeBOp>(linalgOp)) {
    return SmallVector<SmallVector<DimType>>{
        SmallVector<DimType>{DimType::Batch, DimType::M, DimType::K},
        SmallVector<DimType>{DimType::Batch, DimType::N, DimType::K},
        SmallVector<DimType>{DimType::Batch, DimType::M, DimType::N}};
  }
// ...
```

include/gc/Analysis/MatmulConfigAnalysis.h
```diff
@@ -0,0 +1,136 @@
...
+using namespace mlir;
+
+// A mock for the taget information
+// TODO: replace it with upstream hardware description model
+struct SystemDesc {
+  // get runtime OMP_NUM_THREADS
+  uint32_t getNumThreads() {
+    char *numThreads = getenv("OMP_NUM_THREADS");
+    if (numThreads) {
+      return std::stoi(numThreads);
```
ZhennanQin, line 27: Please check if return value is > 0.

include/gc/Analysis/MatmulConfigAnalysis.h
```diff
@@ -0,0 +1,136 @@
...
+    }
+    return 1;
+  }
+  // get cache size by cacheLevel
+  size_t getCacheSize(uint8_t cacheLevel) {
+    if (cacheLevel == 1) {
+      char *cacheSize = getenv("L1_CACHE_SIZE");
+      if (cacheSize) {
+        return std::stoi(cacheSize);
```
ZhennanQin, line 36: Better create a util function to handle unexpected input.

include/gc/Analysis/MatmulConfigAnalysis.h
```diff
@@ -0,0 +1,136 @@
...
+    } else if (cacheLevel == 3) {
+      char *cacheSize = getenv("L3_CACHE_SIZE");
+      if (cacheSize) {
+        return std::stoi(cacheSize);
+      }
+    }
+    return 0;
+  }
+
+  SmallVector<size_t> getContractionOperationMaxVectorLength() {
```
ZhennanQin, line 52: Please add a comment to describe the two values.

include/gc/Analysis/MatmulConfigAnalysis.h
```diff
@@ -0,0 +1,136 @@
...
+    return {512UL, 512UL};
+  }
+};
+
+struct MatmulConfig {
```
ZhennanQin, line 57: Add comments about the meaning of each config.

zhczhong: Added

include/gc/Analysis/MatmulConfigAnalysis.h
zhczhong: Added

include/gc/Analysis/MatmulConfigAnalysis.h
zhczhong: Add a util function `getPositiveIntFromStr` for it

include/gc/Analysis/MatmulConfigAnalysis.h
```diff
@@ -0,0 +1,150 @@
+//===-- MatmulConfigAnalysis.h - DESC ---------------------------*- C++ -*-===//
```
ciyongch, line 1: nit: what's the meaning of `DESC` here?

include/gc/Transforms/Passes.td
```diff
@@ -58,4 +47,30 @@ def LinalgToXeGPU : Pass<"linalg-to-xegpu", "func::FuncOp"> {
...
+}
+
+def SinkOpIntoInnerLoop : Pass<"sink-op-into-inner-loop"> {
+  let summary = "Sink operations into inner loops";
+  let description = [{The pass tries to sink operations into inner loops as deep as possible to maximize the chance for outer loop optimization.
+  }];
+  let dependentDialects = [];
+}
+
+def MergeNestedForall : Pass<"merge-nested-forall"> {
```
ciyongch, line 70: This pass seems quite common, does upstream have equivalent one?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+namespace mlir {
+namespace gc {
+
+#define DEBUG_TYPE "matmul-config-analysis"
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &ss,
+                              const MatmulConfig &config) {
+
+  ss << "MBlock: " << config.MBlock << ", NBlock: " << config.NBlock
```
ciyongch, line 21: Is it better print the configs in the order of loop nest?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+// check if the config is valid
+bool isValidConfig(const MatmulConfig &config, SystemDesc &sysDesc,
+                   ArrayRef<uint32_t> shape) {
+  if (config.innerMostMBlock == 0 || config.innerMostNBlock == 0 ||
+      config.innerMostKBlock == 0) {
+    return false;
+  }
+
+  if (shape[0] % config.innerMostMBlock != 0 ||
```
ciyongch, line 77: Need to check the `shape` size before accessing its element?
And the arg of `sysDesc` is useless here, can be removed?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,438 @@
...
+  }
+  return actualThreads == numThreads;
+}
+
+// calculate the cost of the hardware efficiency(whether the vector register is
+// fully utilized)
+double hardwareEfficiencyCost(linalg::LinalgOp &linalgOp,
+                              ArrayRef<uint32_t> shape,
+                              const MatmulConfig &config, SystemDesc &sysDesc) {
+  llvm::TypeSize dtypeSize = DataLayout().getTypeSizeInBits(
```
ciyongch, line 101: Since it's only considering the vector register utilization, using `vectorRegEfficiencyCost` might be better?
Is it possible that the different in&out have different data type?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,438 @@
...
+      ShapeAdaptor(linalgOp.getDpsInputs()[1].getType()).getElementType());
+  size_t maxVectorLength = sysDesc.getMaxVectorLength() / dtypeSize;
+  double cost = (maxVectorLength - config.innerMostMBlock % maxVectorLength) %
+                    maxVectorLength * 1.0 / config.innerMostMBlock +
```
ciyongch, line 105: Not sure if this is the ideal algo to calculate the register utilization, as the computation of matmul is not always loading full register width from M or K dimension (or it might depends on impl), while most likely it's depending on `N` for the dotproduct? Do you think this cost factor really matters?

include/gc/Analysis/MatmulConfigAnalysis.h
zhczhong: It means "description". I replace it with the concrete description

include/gc/Transforms/Passes.td
zhczhong: It seems that there is no such pass to handle the scf.forall in the upstream(https://mlir.llvm.org/docs/Passes/#scf-dialect-passes). And only scf.parallel canalization could merge several scf.parallel(if no other ops between them).

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: Thanks for the advice! The order has been changed according to the loop order

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: Thanks! The function is unused and I have removed this function

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: The name has changed accordingly. The input and output should have the same data type now and the deepTileMatmul will break `bf16xbf16->bf16` to `bf16xbf16->f32+cast(f32->bf16)` internally.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: The cost model is still under-tuning and this could be further adjusted. This cost is used to let the inner block be as close to the register length or its multiple as possible. It is hard to know how brgemm utilizes register at the linalg level so here just an approximation. And in the current benchmark, this is good for brgemm performance at most cases.

include/gc/Analysis/MatmulConfigAnalysis.h
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }
+  return idxList;
+}
+
+// Get the operand dim type for every operand for the given linalg op
+inline FailureOr<SmallVector<SmallVector<DimType>>>
+getOprandDimType(linalg::LinalgOp &linalgOp) {
+  // TODO: replace the linalgx op with generic op
+  if (isa<linalg::MatmulOp>(linalgOp)) {
```
crazydemo, line 91: align with the following coding style, use `llvm::isa`?

lib/gc/Transforms/DeepTileContractionNamedOp.cpp
```diff
@@ -0,0 +1,1032 @@
...
+        CSlice4 = CSlice3[0, im * iim_block_: (im + 1) * iim_block_, in *
+iin_block_: (in + 1) * iin_block_] (init with 0 when ok == 0)
+        brgemm(bs=KNumInnerBlock, M=iim_block_, N=iin_block_, K=iik_block,
+A=ASlice3, B=BSlice3, C=CSlice4, onlyUpdate=(ok!=0));
+      }
+    }
+}
+C = final_reduce(CSlice)
+*/
+struct deepTileMatmul : public OpInterfaceRewritePattern<linalg::LinalgOp> {
```
crazydemo, line 538: It is better to change the name of `struct / class` into `DeepTileMatmul` style to align with the llvm coding style.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
ciyongch: I'm considering the AMX register, for BF16 datatype, this cost function prefers multiple of (32,32,32), while multiple of (64,64,64) for INT8 datatype, I assume 32,64,32 would also an option for INT8. Let's revise this with more benchmark result.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,422 @@
...
+  return 1 / computationIntensity;
+}
+
+using CostModelFn =
+    std::function<float(linalg::LinalgOp &linalgOp, ArrayRef<uint32_t> shape,
+                        MatmulConfig cfg, SystemDesc &sysDesc)>;
+
+// filter the config by the cost model
+std::vector<MatmulConfig>
+filterConfigByCostModel(std::vector<MatmulConfig> configs,
```
ciyongch, line 163: Using const reference to avoid extra copy or mutation?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,422 @@
...
+  std::vector<uint32_t> NThreadsCandidates =
+      getCandidate((uint32_t)threads, 1U);
+  std::vector<uint32_t> KThreadsCandidates =
+      getCandidate((uint32_t)threads, 1U);
+  uint32_t noSmallBlockNeedThreshold = 8 * 8U;
+  std::vector<uint32_t> MBlockCandidates = getCandidate(
+      (uint32_t)shape[0], shape[0] > noSmallBlockNeedThreshold ? 8U : 1U,
+      (uint32_t)shape[0]);
+  std::vector<uint32_t> NBlockCandidates =
+      getCandidate((uint32_t)shape[1],
```
ciyongch, line 212: Check the size of `shape` before indexing it.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,422 @@
...
+                   shape[1] > noSmallBlockNeedThreshold ? 8U : 1U, shape[1]);
+  std::vector<uint32_t> KBlockCandidates =
+      getCandidate((uint32_t)shape[2],
+                   shape[2] > noSmallBlockNeedThreshold ? 8U : 1U, shape[2]);
+  std::vector<uint32_t> innerMostMBlockCandidates = getCandidate(
+      (uint32_t)shape[0], shape[0] > noSmallBlockNeedThreshold ? 8U : 1U, 256U);
```
ciyongch, line 218: `>` -> `>=`?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,422 @@
...
+  std::vector<uint32_t> innerMostNBlockCandidates = getCandidate(
+      (uint32_t)shape[1], shape[1] > noSmallBlockNeedThreshold ? 8U : 1U, 256U);
+  std::vector<uint32_t> innerMostKBlockCandidates = getCandidate(
+      (uint32_t)shape[2], shape[2] > noSmallBlockNeedThreshold ? 8U : 1U, 256U);
+  if (givenInnermostBlock.size() == 3) {
+    innerMostMBlockCandidates =
```
ciyongch, line 224: Better to change the logic to only calculate the candidates when `givenInnermostBlock.size() != 3` or its value is not 0?

lib/gc/Transforms/MergeNestedForall.cpp
```diff
@@ -0,0 +1,94 @@
...
+struct MergeNestedForallLoops : public OpRewritePattern<scf::ForallOp> {
+  using OpRewritePattern<scf::ForallOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(scf::ForallOp op,
+                                PatternRewriter &rewriter) const override {
+    Block &outerBody = *op.getBody();
+    if (!llvm::hasSingleElement(outerBody.without_terminator()))
+      return failure();
+
+    scf::ForallOp innerOp = dyn_cast<scf::ForallOp>(outerBody.front());
```
ciyongch, line 28: Seems it only supports two level `forall` merge?

lib/gc/Transforms/Pipeline.cpp
```diff
@@ -32,6 +32,17 @@
 
 namespace mlir::gc {
 
+void populateCleanUpPasses(mlir::OpPassManager &pm) {
```
ciyongch, line 35: Why do we need LICM/mem2reg/topologicalSort pass inside cleanup stage?

lib/gc/Transforms/TilingUtil.hpp
```diff
@@ -0,0 +1,26 @@
...
+#define TEMPORARY_TILEUSINGINTERFACE_X_H
+
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Linalg/Transforms/Transforms.h"
+#include "mlir/Interfaces/TilingInterface.h"
+#include <optional>
+namespace mlir {
+namespace linalgX {
+
+FailureOr<linalg::ForallReductionTilingResult> tileReductionUsingForall(
```
ciyongch, line 19: What's the difference between this vs upstream one?

zhczhong: The upstream one could only support do the K slicing for the linalg op with only one reduction iterator and not support do it for blocking layout with two reduction iterators(MKmk). Here is an enhancement for that.

lib/gc/Transforms/MergeNestedForall.cpp
zhczhong: This is a rewrite pattern and will merge recursively merge any two neighbor forall op if possible. For example
```
forall1
forall2
forall3
---------->
forall1+2
forall3
---------->
forall1+2+3
```

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,422 @@
...
+  return configs;
+}
+
+// read the config from the attributes for tuning
+bool readConfigFromAttrs(MatmulConfig &config, ArrayRef<NamedAttribute> attrs) {
+  bool hasPredefinedConfig = false;
+  for (auto &attr : attrs) {
+    if (attr.getName() == "KBlock") {
+      config.KBlock = cast<IntegerAttr>(attr.getValue()).getInt();
+      hasPredefinedConfig = true;
```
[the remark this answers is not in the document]
zhczhong: Currently, it only supports reading full config from attrs. Will add a check here

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,422 @@
...
+    innerMostNBlockCandidates =
+        givenInnermostBlock[1] != 0
+            ? std::vector<uint32_t>{givenInnermostBlock[1]}
+            : innerMostNBlockCandidates;
+    innerMostKBlockCandidates =
+        givenInnermostBlock[2] != 0
+            ? std::vector<uint32_t>{givenInnermostBlock[2]}
+            : innerMostKBlockCandidates;
+  }
+  for (uint32_t MThreads : MThreadsCandidates) {
```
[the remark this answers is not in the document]
zhczhong: Thanks for the advice. For single matmul testing, it doesn't consume too much time and the total compile time is around ~200ms. Will add a todo here for further improvement

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,422 @@
...
+               << "M: " << M << ", N: " << N << ", K: " << K << "\n");
+
+    // try to read the config from the attributes
+    SmallVector<NamedAttribute> attrs(linalgOp->getAttrs());
+    bool hasPredefinedConfig = readConfigFromAttrs(config, attrs);
+
+    // if there is a given config, skip the cost model
+    if (!hasPredefinedConfig) {
+      LLVM_DEBUG(llvm::dbgs() << "No predefined config\n");
+      SmallVector<std::tuple<CostModelFn, std::string, double>> costModelList =
```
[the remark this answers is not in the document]
zhczhong: Thanks for the advice! The priority now is decided by the execution order of these cost models(the last one has the highest priority). Add a todo here to explore different weight and priority in the future.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: Changed

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: changed

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: Add a check here

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,422 @@
...
+  // factor
+  std::vector<uint32_t> candidates;
+  for (uint32_t i = 1; i <= num; i++) {
+    if (num % i == 0 && i <= ceil && i >= floor) {
+      candidates.push_back(i);
+    }
+  }
+  // the pow of 2
+  uint32_t candidate = 1U;
+  while (candidate < num && candidate <= ceil && candidate >= floor) {
```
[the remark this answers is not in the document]
zhczhong: Thanks for pointing out! It should be fixed now

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: Thanks! Replace it with arrayref

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,422 @@
...
+}
+
+constexpr unsigned bitPerByte = 8;
+// calculate the cost of the memory consumption on the thread
+double memoryConsumptionOnThreadCost(linalg::LinalgOp &linalgOp,
+                                     ArrayRef<uint32_t> shape,
+                                     const MatmulConfig &config,
+                                     SystemDesc &sysDesc) {
+  uint32_t M = shape[0], N = shape[1], K = shape[2];
+  llvm::TypeSize dtypeSize = DataLayout().getTypeSizeInBits(
```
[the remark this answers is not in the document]
zhczhong: Thanks! Have replaced it with `getTypeSize()`

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,422 @@
...
+                (maxVectorLength - config.innerMostNBlock % maxVectorLength) %
+                    maxVectorLength * 1.0 / config.innerMostNBlock;
+  return cost;
+}
+
+// calculate the cost of the workload balance
+double workloadBalancedCost(linalg::LinalgOp &linalgOp,
+                            ArrayRef<uint32_t> shape,
+                            const MatmulConfig &config, SystemDesc &sysDesc) {
+  uint32_t M = shape[0], N = shape[1], K = shape[2];
```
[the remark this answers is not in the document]
zhczhong: Add a check here

lib/gc/Transforms/DeepTileContractionNamedOp.cpp
zhczhong: Thanks for the advice! The name has been changed

include/gc/Analysis/MatmulConfigAnalysis.h
zhczhong: Changed

lib/gc/Transforms/Pipeline.cpp
zhczhong: Has moved the LICM/mem2reg out of clean up stage and only call them when needed

lib/gc/Transforms/DeepTileContractionNamedOp.cpp
```diff
@@ -0,0 +1,1032 @@
...
+
+// Find the parent fill op of a value and will penetrate pack/pad ops
+static Operation *findParentFillOp(Value val) {
+  SmallVector<StringRef> skipOpList = {"tensor.pack", "tensor.pad"};
+  Operation *currentOp = val.getDefiningOp();
+  while (currentOp &&
+         llvm::find(skipOpList, currentOp->getName().getStringRef()) !=
+             skipOpList.end() &&
+         !isa<linalg::FillOp>(currentOp)) {
+    currentOp = currentOp->getResult(0).getDefiningOp();
```
yifeizh2, line 246: There it will goes into an endless loop, shall be `currentOp->getOperand(0).getDefiningOp()` instead?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: Yes, a smaller NBlock and MBlock are also enough for amx case. I add a todo here to further adjust this cost model.

lib/gc/Transforms/DeepTileContractionNamedOp.cpp
zhczhong: Thanks for pointing it out! I replace it with `getOperand`

zhczhong:
The target description is replaced with querying DLTI information and a corresponding test is also added

include/gc/Analysis/MatmulConfigAnalysis.h
```diff
@@ -10,64 +10,88 @@
...
+      }
     }
     return 0;
   }
 
   // get the maximum vector length in bits
   size_t getMaxVectorLength() {
-    char *maxVectorLanes = getenv("MAX_VECTOR_LENGTH");
-    return getPositiveIntFromStr(maxVectorLanes, 512);
+    DataLayout layout = DataLayout(module);
```
ciyongch, line 67: Seems we can also add `layout` as a private member and only create once at instantiation stage.

lib/gc/Transforms/TilingUtil.hpp
ciyongch: Thanks, better to add a notes for the difference here.

include/gc/Analysis/MatmulConfigAnalysis.h
zhczhong: Thanks for the advice! Has replaced `moduleOp` with `layout` and `context` as private member

lib/gc/Transforms/TilingUtil.hpp
zhczhong: Added

lib/gc/Transforms/MergeNestedForall.cpp
ciyongch: I see, thanks.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
ciyongch: forgot?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,443 @@
...
+      cfgItemCnt++;
+    } else if (attr.getName() == "innermostNBlock") {
+      config.innerMostNBlock = cast<IntegerAttr>(attr.getValue()).getInt();
+      cfgItemCnt++;
+    } else if (attr.getName() == "innermostKBlock") {
+      config.innerMostKBlock = cast<IntegerAttr>(attr.getValue()).getInt();
+      cfgItemCnt++;
+    }
+  }
+  return cfgItemCnt == 9;
```
ciyongch, line 326: I was referring to validate all the fields of `config`, in case user provides any invalid number to any field of it by mistake.

zhczhong: Thanks for the reminder! The check for every field and the check for index access are added.

include/gc/Analysis/MatmulConfigAnalysis.h
```diff
@@ -0,0 +1,168 @@
...
+#include "mlir/Interfaces/DataLayoutInterfaces.h"
+
+namespace mlir {
+namespace gc {
+
+using namespace mlir;
+
+struct SystemDesc {
+  // get runtime OMP_NUM_THREADS
+  uint32_t getNumThreads() {
```
Yun-Fly, line 24: Could we move this struct to another common header for better reuse?

ZhennanQin: Let's do this when implementing centerized DLTI. You can duplicate this to avoid dependency.

zhczhong: Yes, I will do it with another PR for this issue https://github.com/intel/graph-compiler/issues/188 and add more verifer/default value.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,468 @@
...
+                         MatmulConfig cfg, SystemDesc &sysDesc)>;
+
+// filter the config by the cost model
+std::vector<MatmulConfig>
+filterConfigByCostModel(ArrayRef<MatmulConfig> configs,
+                        linalg::LinalgOp &linalgOp, ArrayRef<uint32_t> shape,
+                        SystemDesc &sysDesc, const CostModelFn &costModel,
+                        float eliminationRatio = 0.5, float threshold = -1) {
+  std::vector<MatmulConfig> result;
+  std::vector<float> costs;
```
Yun-Fly, line 179: What is the meaning of `eliminationRatio` here?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,468 @@
...
+           {computationIntensityOnL2Cache, "computationIntensityOnL2Cache", -1},
+           {memoryConsumptionOnThreadCost, "memoryConsumptionOnThreadCost",
+            -1}};
+      SmallVector<uint32_t> shape = {M, N, K};
+      std::vector<MatmulConfig> configCandidates =
+          prepareConfigCandidates(root, sysDesc, shape, givenInnermostBlock);
+      for (auto [fn, name, threshold] : costModelList) {
+        configCandidates = filterConfigByCostModel(
+            configCandidates, linalgOp, shape, sysDesc, fn, 0.5, threshold);
+      }
```
Yun-Fly, line 456: I have a little concern about this multiple level sorting. Although you have marked execution order as priority. However, in some cases, the result of these cost model pipeline maybe just identical with the single execution of last cost model. In another word, any other cost model before last one would make no sense in fact. IMO, the lower priority cost model should only be applied for each two candidates, whose results from higher priority cost model are equal.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -0,0 +1,468 @@
...
+template <typename T>
+static llvm::raw_ostream &operator<<(llvm::raw_ostream &ss,
+                                     std::vector<T> arry) {
+  ss << "[";
+  for (auto [idx, a] : llvm::enumerate(arry)) {
+    if (idx != 0) {
+      ss << ", ";
+    }
+    ss << a;
+  }
```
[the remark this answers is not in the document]
zhczhong: Changed

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: Change it `preserveRatio` to reduce the confusion. It means that it only preserves the top `preserveRatio`(30% for example) candidates and the rest will be thrown away.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: Yes, this is why I introduce the `preserveRatio` and `threshold`. This will filter out the bad values on the first several cost models. The rest of the candidates are at least not bad for the low-priority models.

Yun-Fly: Ok, I see. But it may confuse me more about the definition of high and low priority, because the low-priority model has already filter out some candidates before high-priority model. What if the high model actually prefer those bad candidates in the eyes of lower model?

zhczhong: Yes, the word `priority` is a little confusing indeed. In fact, every cost model will make their own choice and filter out the unacceptable config. The last cost model will pick the best from the rest candidates(or provide the candidates list if needed). In your case, there is a contradiction among these cost models. We could only balance them and pick a config that is not too bad for every model.

Yun-Fly: That's fine. So the order of cost model list may also affect final config? Do we need to treat it as a tuning option in the future?

test/gc/Transform/deepTileContractionNamedOp.mlir
```diff
@@ -0,0 +1,157 @@
...
+    // CHECK: tensor.empty() : tensor<8x8x32x32xf32>
+    // CHECK: scf.for
+    // CHECK: scf.for
+    // CHECK: tensor.extract_slice
+    // CHECK: tensor.extract_slice
+    // CHECK: scf.for
+    // CHECK: tensor.extract_slice
+    // CHECK: tensor.extract_slice
+    // CHECK: tensor.extract_slice
+    // CHECK: tensor.extract_slice
```
Yun-Fly, line 61: Is this FileCheck too relax to check missing information, saying `OFFSET` and `TILESIZE`?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: Yes, the cost model and the execution order will be further tuned in the future

test/gc/Transform/deepTileContractionNamedOp.mlir
zhczhong: Thanks for the reminder! I add more check in the test

cmake/llvm-version.txt
```diff
@@ -1 +1 @@
-89946bda5e1c7ceaf6d26634cc8c8c9498d9f7be
+fbee8d5a65defae402f98794d778fe724d24cc51
```
ciyongch, line 1: Shall we update llvm in a separate PR?

zhczhong: Yes, I have reverted this change and will use a new PR if needed

lib/gc/Transforms/DeepTileContractionNamedOp.cpp
```diff
@@ -0,0 +1,1005 @@
...
+  Operation *linalgOp = nullptr;
+  Operation *castOp = nullptr;
+};
+
+// Split a low precision matmul(bf16xbf16->bf16) to a combination
+// matmul(bf16xbf16->f32) + cast(f32->bf16)
+// if needFurtherFuse=true, a middle temporary linalgOp(bf16xbf16->(f32,bf16))
+// will be created
+static FailureOr<DtypeLegalizeResult>
+matmulDtypeLegalize(RewriterBase &rewriter, Operation *op,
```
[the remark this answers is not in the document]
zhczhong: It is about the accumulation type. Matmul(bf16xbf16->bf16) will do accumulation on bf16 while matmul(bf16xbf16->f32)+cast(f32->bf16) will do accumulation on f32 and cast back to bf16.

lib/gc/Transforms/DeepTileContractionNamedOp.cpp
```diff
@@ -0,0 +1,1005 @@
...
+                    bool needCopyInit = true, bool needFurtherFuse = false) {
+  linalg::LinalgOp linalgOp = dyn_cast<linalg::LinalgOp>(op);
+  Location loc = linalgOp->getLoc();
```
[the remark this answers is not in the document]
zhczhong: Have moved it after check

lib/gc/Transforms/DeepTileContractionNamedOp.cpp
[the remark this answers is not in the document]
ciyongch: I was referring to having different in & out datatype for matmul ops, so that the fusion will handle the case like  `linalg.matmul(bf16:bf16:f32) + tensor.cast()` instead of an internal pass to handle this?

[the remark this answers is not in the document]
zhczhong: The current fusion cannot fuse the cast into the innermost loop if the K/reduction axis is not full. So the `deepTileMatmul` will temporarily fuse the cast into the innermost loop in a template way.

```
deep tile matmul:
for(M : MBlocks) {
    for(K: KBlocks) {
        for(N: NBlocks) {
            C = brgemm(M, K, N);
            cast_to_bf16(C);
        }
    }
}
current fusion: 
for(M : MBlocks) {
    for(K: KBlocks) {
        for(N: NBlocks) {
            C = brgemm(M, K, N);
        }
    }
    cast_to_bf16(bigger_C);
}
```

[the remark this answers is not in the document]
ciyongch: I see, so this is one of the inefficiency of automatically fusion approach. Thanks for the explaination.

ZhennanQin (approved):

Yun-Fly (approved):

ciyongch (approved):

PR #94, opened 2024-05-22
fix clang-tidy.yml

WangJialei-A:
Install OpenMP to avoid cmake check fail

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -14,7 +14,7 @@ jobs:
 
     steps:
     - name: Install OpenMP
-      run: "apt install -y libomp-dev"
+      run: "sudo apt install -y libomp-dev"
       
     - name: Fetch sources
       uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


leshikus:
1. I see there is an inconsistency with quotes. You use them for 'run' clause and don't use for 'name' clause.

2. There is a major problem here. Why this `clang-tidy` check was implemented in the first place (without using self-hosted runners or re-using the build to collect symbols)? Is there any CI roadmap which requests replacing `clang-format` with `clang-tidy`. This all looks strange for me because one week I implement `clang-format` and it is replaced. I don't say `clang-tidy` is bad, it just was not discussed (at least with me) before it was implemented.

leshikus:
Well, at least I expect to see some motivation for any CI feature. Is it nice output? Is it for the deep static analysis?

leshikus:
I've double checked the initial PR. Why do we follow requirements for llvm project? Is this project supposed to be a part of llvm?

leshikus (approved):
There were complains that `clang-tidy` works slowly. For me it works quickly. Maybe then we can postpone questions about optimizing it with self-hosted runner or re-using symbols till we have a problem in hand.

kurapov-peter:
@leshikus, we want the static analysis, yes. No, it is not supposed to be a part of llvm. As to all the other questions, please decide what's best, I trust your judgment.

PR #96, opened 2024-05-22
[Build script] Fix debug messages

Devjiu:
Remove debug messages. There are still possible misuse of script as options without "-" in the start will be ignored.

### DIFF
--- a/scripts/compile.sh
+++ b/scripts/compile.sh
@@ -3,9 +3,9 @@
 repo=intel/graph-compiler
 
 # Uncomment for script debug
-# set -ex
+# set -x
 
-function print_usage() {
+print_usage() {
     echo "Usage:"
     echo "$0 "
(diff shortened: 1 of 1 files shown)


leshikus (commented):
Not using `-e` is a bad practice; moreover, `-e` should be set at the script start with `set -e`

I don't like removing it

If you have a command `cmd`, which could potentially fail, and you want to handle that, use `if cmd` or `cmd ||`

Devjiu:
> Not using `-e` is a bad practice; moreover, `-e` should be set at the script start with `set -e`
> 
> I don't like removing it
> 
> If you have a command `cmd`, which could potentially fail, and you want to handle that, use `if cmd` or `cmd ||`

I've reworked this part based on Andrey's comments, now the options parser is a little simpler and I've removed the bash syntax since it's not needed yet.

AndreyPavlenko (approved):

PR #97, opened 2024-05-23
Add all-in-one driver for compile-and-execute

Menooker:
For usage see unittest.

Also add a wrapper for `mlir::ExecutionEngine` for further developing of constant cache.


lib/gc/ExecutionEngine/CPURuntime/ConstantCache.cpp
```diff
@@ -0,0 +1,40 @@
...
+  ref.basePtr = (char *)base->get_buffer_unsafe() + offset;
+  ref.data = ref.basePtr;
+  ref.offset = 0;
+  memset(ref.sizes, 0, sizeof(ref.sizes));
+  memset(ref.strides, 0, sizeof(ref.strides));
+}
+
+static std::unordered_map<uint64_t, std::shared_ptr<cached_graph_tensor>> cache;
+
+std::shared_ptr<cached_graph_tensor> query_cached_tensor(uint64_t key) {
```
Menooker, line 23: This is a dummy interface of cache manager, just for testing when const cache is not available. @niuxiaog Feel free to remove/modify the cache manager interfaces here.

lib/gc/ExecutionEngine/JitWrapper/Module.cpp
```diff
@@ -0,0 +1,254 @@
...
+#include "llvm/Support/InitLLVM.h"
+#include "llvm/Support/TargetSelect.h"
+
+#define likely(x) __builtin_expect(!!(x), 1)
+#define unlikely(x) __builtin_expect(!!(x), 0)
+
+namespace mlir {
+namespace gc {
+
+static DialectRegistry initDialects() {
```
Menooker, line 27: A unified place to get the dialect registry, and initialize LLVM/MLIR library.

unittests/ExecutionEngine/JitWrapper.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include "llvm/Support/ErrorOr.h"
+#include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Support/SourceMgr.h"
+#include "llvm/Support/raw_ostream.h"
+#include "gtest/gtest.h"
+#include <memory>
+
+using namespace mlir;
+
+static const char code1[] = R"mlir(
```
Menooker, line 26: This is example IR for no-folding program.

unittests/ExecutionEngine/JitWrapper.cpp
```diff
@@ -0,0 +1,175 @@
...
+  OwningMemRef<float, 1> bufB{
+      {128}, {128}, [](float &ptr, ArrayRef<int64_t> idx) { ptr = idx[0]; }};
+  OwningMemRef<float, 1> bufC{{128}, {128}};
+  void *args[] = {&*bufA, &*bufB, &*bufC};
+  jited.get()->call(args);
+  for (int i = 0; i < 128; i++) {
+    ASSERT_EQ(bufC[{i}], 1.0f + i);
+  }
+}
+
```
Menooker, line 71: This is an example for fold+compute program. The pass should export `__num_orig_num_args= original number of inputs+outputs`. In this example, `abc` as input and `d` as output.

` __fold_buffer_ids` as an array of `uint64`. The first element is the size of the array (not counting itself). The remaining elements are the `bufferIDs` for each cached buffer in cache manager (the "key").  In this example, we manually register the keys `114514, 1919810` in 

```c++
  ASSERT_TRUE(gc::reg_cached_tensor(114514, proxy, 0));
  ASSERT_TRUE(gc::reg_cached_tensor(1919810, proxy, 128 * sizeof(float)));
```

Menooker: `__fold_args` are the `bufferIndex` for arguments of `fold()` function. Here below, we defined `fold(a,b, foldedA, foldedB)` and we have `bufIdx: a=0, b=1, c=2, d=3, foldedA=4, foldedB=5`. So `__fold_args = [0,1,4,5]` and we add the length of the array at the head of the array.

unittests/ExecutionEngine/JitWrapper.cpp
```diff
@@ -0,0 +1,175 @@
...
+llvm.mlir.global constant @__num_orig_num_args(4 : i32) : i32
+llvm.mlir.global constant @__fold_buffer_ids(dense<[2, 114514, 1919810]> : tensor<3 x i64>) : !llvm.array<3 x i64>
+// a,b, foldedA,foldedB
+llvm.mlir.global constant @__fold_args(dense<[4, 0, 1, 4, 5]> : tensor<5xi32>) : !llvm.array<5 x i32>
+// foldedA, foldedB, c, d
+llvm.mlir.global constant @__compute_args(dense<[4, 4, 5, 2, 3]> : tensor<5xi32>) : !llvm.array<5 x i32>
+
+func.func @fold(%a: tensor<128xf32>, %b: tensor<128xf32>) -> (tensor<128xf32>, tensor<128xf32>) attributes { llvm.emit_c_interface } {
+    %c0 = arith.constant 0 : index
+    cpuruntime.printf "HI%zu\n" %c0 : index
```
Menooker, line 85: We print "HI0" in `fold()`, and we check if there is "HI0" in stdout for 1st run, 2nd run, and 3rd run (after evicting cache)

niuxiaog: Because `%c0 = arith.constant 0 : index` and `cpuruntime.printf "HI%zu\n" %c0 : index` are not related to any constant input, I think they should **not** be processed by pass `constant-weights-folding`.

lib/gc/ExecutionEngine/CPURuntime/ConstantCache.cpp
niuxiaog: My implementation is at https://github.com/intel/graph-compiler/commit/5eb0ac014cd3fdf48a27eee05c71bb8aed719274. Let's test it when possible.

AndreyPavlenko:
I think it would be nice to use the same code style for the entire code. Currently, one part of the code is in the CamelCase, another - in snake_case.

unittests/ExecutionEngine/JitWrapper.cpp
Menooker: This is the IR **after** `constant-weights-folding`. I used `printf` to find if the `fold()` code is executed or not.

include/gc/ExecutionEngine/CPURuntime/ConstantCache.hpp
```diff
@@ -0,0 +1,137 @@
+#ifndef GC_EXECUTIONENGINE_CPURUNTIME_CONSTANT_CACHE_H
```
ZhennanQin, line 1: License header here.

Menooker: Sure. This PR is WIP for preview, since it is has pending dependency. I will update the code when it is finalized.

WangJialei-A: @Menooker 
Please rebase `yijie/pipeline` and `yijie/mainfunc_wrapper` to enable the license check workflow

unittests/ExecutionEngine/JitWrapper.cpp
niuxiaog: The problem is that whether a constant operation which is not related to any constant input should be moved to `fold`. Currently I leave it in `compute`.

Menooker: Here the code is the manual written IR for testing, not the real output of  `constant-weights-folding`. If I use the real output `constant-weights-folding`, how can I test if the `fold()` func is skipped at the second execution?

niuxiaog: OK I got you. Let's use this manually written example as a unit test.

Menooker:
> I think it would be nice to use the same code style for the entire code. Currently, one part of the code is in the CamelCase, another - in snake_case.

I agree with you. The code was copied from legacy. Now I have updated the coding style to CamelCase.

lib/gc/ExecutionEngine/Driver/Driver.cpp
```diff
@@ -0,0 +1,254 @@
...
+  }
+  JitModuleFuncT compute;
+  {
+    auto expectCompute = engine->lookupPacked(defaultComputeName);
+    if (!expectCompute) {
+      return expectCompute.takeError();
+    }
+    compute = *expectCompute;
+  }
+  llvm::ArrayRef<uint64_t> foldBufferIds;
```
niuxiaog, line 82: Better to unify the element dtypes here with the Globals in MLIR module. "__num_orig_num_args", "__fold_args" and "__compute_args" are i32, and "__fold_buffer_ids" is i64.

Menooker: __fold_buffer_ids is the UID of a buffer, so we use i64. `__num_orig_num_args` and others are i32 because we won't meet too many args in the same function if it is reasonable. We use i32 for `__fold_args` and `__compute_args` for performance because it save half of space and cache (!!) at each execution time.

unittests/CMakeLists.txt
AndreyPavlenko, line 1: Do we need another one tests folder?

Menooker: The folder `tests` is for MLIR-related FileCheck-based tests and `unittests` is for MLIR-related C++ gtests. The convention of `tests` and `unittests` is borrowed from MLIR upstream - they use these folders too.
I have several pending PRs to put the C++-based gtests in `unittests` folder.

Menooker:
Hi, all! I have removed the constant cache related code. This PR now only contains all-in-one `ModuleOp` to `JitModule` conversion. And it also provides a unified place to initialize the compiler (LLVM JIT, MLIR passes) and DialectRegistry.

AndreyPavlenko (approved):

LongshengDu (approved):

lib/gc/ExecutionEngine/Driver/Driver.cpp
```diff
@@ -0,0 +1,82 @@
...
+const DialectRegistry &initCompilerAndGetDialects() {
+  static DialectRegistry reg = initDialects();
+  return reg;
+}
+
+static const char defaultComputeName[] = "_mlir_ciface_compute";
+
+llvm::Expected<std::shared_ptr<JitModule>>
+JitModule::create(Operation *op, const DriverOptions &options) {
+  if (options.runTransforms) {
```
ZhennanQin, line 49: if `runTransforms` is false, I guess the llvm codegen won't work directly on top of oneDNN graph or Linalg dialect. It means runTransforms is mandatory. Let's remove it until we have clear design about the optimization level of the CPU pipeline.

include/gc/ExecutionEngine/Driver/Driver.h
```diff
@@ -0,0 +1,69 @@
...
+    // inputs
+    llvm::SmallVector<void *, 32> realargs;
+    realargs.reserve(numArgs);
+    for (size_t i = 0; i < numArgs; i++) {
+      realargs.push_back(&args[i]);
+    }
+    compute(realargs.data());
+  }
+
+  // directly call compute(). args should be an array of void*. args[i] should
```
ZhennanQin, line 52: // -> ///

include/gc/ExecutionEngine/Driver/Driver.h
```diff
@@ -0,0 +1,69 @@
...
+class JitModule {
+public:
+  static llvm::Expected<std::shared_ptr<JitModule>>
+  create(Operation *op, const DriverOptions &options = {});
+
+  // args should be an array of XXXMemrefType*
+  void call(GeneralMemrefPtr *args, std::size_t numArgs) {
+    // Silly code, MLIR execution engine requires pointers of real args as
+    // inputs
+    llvm::SmallVector<void *, 32> realargs;
```
ZhennanQin, line 44: Better to cache this vector.

include/gc/ExecutionEngine/Driver/Driver.h
```diff
@@ -0,0 +1,69 @@
...
+  // be a pointer to the real data. For passing memref, users need to 1) create
+  // a pointer to XXXMemrefType 2) store the pointer to pointer to XXXMemrefType
+  // in args[i]
+  void callRaw(void **args) { compute(args); }
+
+  JitModule(std::unique_ptr<ExecutionEngine> engine, JitModuleFuncT compute);
+  ~JitModule();
+
+private:
+  std::unique_ptr<ExecutionEngine> engine;
```
ZhennanQin, line 62: I don't see it's used anywhere.

ZhennanQin: Ah, just found it.

lib/gc/ExecutionEngine/Driver/Driver.cpp
Menooker: If the input IR is in LLVM-dialect or ready for LLVM-codegen, it does not need to run the passes. This is not related to opt level.

include/gc/ExecutionEngine/Driver/Driver.h
Menooker: We can't cache it. The args may change and we need to consider multi-threading.

lib/gc/ExecutionEngine/Driver/Driver.cpp
ZhennanQin: I see. Let's keep it then.

include/gc/ExecutionEngine/Driver/Driver.h
ZhennanQin: Just cache the vector to avoid creating it many times. We still need to set the pointer into it for each execution.

ZhennanQin: Let's optimize this part later when it becomes a performance bottleneck.

Menooker: What if multiple threads are calling `JitWrapper::call`? `llvm::SmallVector<void *, 32> realargs ` allocates on stack in the happy path, so usually we are using stack memory which will be fast. We can re-visit it here, to use `alloca` intrinsics of C++ compiler here.

ZhennanQin (approved):

Issue #57, opened 2024-05-09
python binding and user-friendly API design


ZhennanQin:
- Support graph programming
- Support benchmark
- Support tuning

aregm:
Please ensure that the decisions are communicated properly.

PR #102, opened 2024-05-27
[Binding] enable binding of cpuruntime dialect

xurui1995:
enable binding of cpuruntime dialect
trace: https://github.com/intel/graph-compiler/issues/57


kurapov-peter (commented):
What are we exposing it for?

xurui1995:
> What are we exposing it for?

check this https://github.com/intel/graph-compiler/issues/57

Menooker:
> > check this #57
> 
> What exactly do you want me to check? :) That doesn't answer the question.

We plan to add py binding for every dialect and pass that we added in MLIR-GC repo, to support direct programming MLIR with python.

PR #103, opened 2024-05-27
[script] fix license format for py files

xurui1995:
fix license format for py files


xurui1995:
see the license in [llvm python file](https://github.com/llvm/llvm-project/blob/main/mlir/python/mlir/ir.py)

Menooker (approved):

WangJialei-A (approved):

PR #107, opened 2024-05-29
[Dialect] [OneDNNGraph] Add ops lowering for llama2 mlp

LongshengDu:
Add llama2 mlp ops lowering, update matmul support for batch broadcast, matmul lowering for 3dx2d flatten. A llama2 mlp graph has been placed in [onednn-graph-llama2.mlir](https://github.com/intel/graph-compiler/blob/longsheng/llma2_onednn_lower/test/gc/Dialect/OneDNNGraph/onednn-graph-llama2.mlir) for codegen testing.

Note: if using conda, first `export LD_PRELOAD=path/to/libomp.so`, then `gc-opt %s --gc-cpu-pipeline | gc-cpu-runner -e main -entry-point-result=void` to execute generated binary. 

Tracking: #117


LongshengDu:
A llama2 mlp graph has been placed in [onednn-graph-llama2.mlir](https://github.com/intel/graph-compiler/blob/longsheng/llma2_onednn_lower/test/gc/Dialect/OneDNNGraph/onednn-graph-llama2.mlir) for codegen testing.

test/gc/Dialect/OneDNNGraph/onednn-graph-to-linalg.mlir
```diff
@@ -24,6 +24,22 @@ func.func @matmul_bias(%arg0: tensor<128x512xbf16>, %arg1: tensor<512x256xbf16>,
   return %0 : tensor<128x256xbf16>
 }
 
+// CHECK-LABEL: @matmul_batch_flatten
+func.func @matmul_batch_flatten(%arg0: tensor<6x128x512xbf16>, %arg1: tensor<512x256xbf16>) -> tensor<6x128x256xbf16> {
```
ciyongch, line 28: Add some cases to cover >3D shapes?

test/gc/Dialect/OneDNNGraph/onednn-graph-ops.mlir
```diff
@@ -11,6 +11,15 @@ func.func @matmul(%arg0: tensor<128x512xbf16>, %arg1: tensor<512x64xbf16>,
   return %0 : tensor<128x64xbf16>
 }
 
+// CHECK-LABEL: @matmul_broadcast_batch
+func.func @matmul_broadcast_batch(%arg0: tensor<6x128x512xbf16>, %arg1: tensor<4x1x512x64xbf16>, 
+                  %arg2: tensor<64xbf16>) -> tensor<4x6x128x64xbf16> {
+  // CHECK: onednn_graph.matmul 
+  %0 = onednn_graph.matmul %arg0, %arg1, %arg2
```
ciyongch, line 18: Add concrete shape check for such special case?

lib/gc/Transforms/Pipeline.cpp
```diff
@@ -92,12 +93,12 @@ void populateCPURuntimePasses(mlir::PassManager &pm) {
 }
 
 void populateLoweringToLLVMPasses(mlir::PassManager &pm) {
+  pm.addPass(createFinalizeMemRefToLLVMConversionPass());
```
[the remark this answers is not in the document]
LongshengDu: We need to FinalizeMemrefToLLVM before SCFToCFPass, other wise will casue `error: 'memref.alloca_scope' op expects region #0 to have 0 or 1 blocks`

test/gc/Dialect/OneDNNGraph/onednn-graph-ops.mlir
LongshengDu: This will check for shape match error, if the tensor shape does not match the op, error will be produced and test will fail. onednn_graph.matmul will only check for shape compatibility, and insert broadcast when lowering.

lib/gc/Transforms/Pipeline.cpp
```diff
@@ -143,6 +144,7 @@ class GCCPUPipeline : public impl::GCCPUPipelineBase<GCCPUPipeline> {
     auto op = getOperation();
     PassManager pm{op->getContext()};
     populateCPUPipeline(pm);
+    // pm.enableIRPrinting();
```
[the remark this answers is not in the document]
LongshengDu: This function also requires disable threading, will try to come up with a better solution for a option to print internal passes later.

test/gc/Dialect/OneDNNGraph/onednn-graph-to-linalg.mlir
LongshengDu: added

lib/gc/Dialect/OneDNNGraph/OneDNNGraphOps.cpp
```diff
@@ -181,22 +201,36 @@ LogicalResult onednn_graph::MatMulOp::inferReturnTypeComponents(
...
     return success();
   };
+  // get broadcasted batch dims from 2 multi-batch mat shape,
+  auto extractBroadcastBatch = [](ShapedType lhsType, ShapedType rhsType,
+                                  SmallVector<int64_t> &outDims) {
+    SmallVector<int64_t> lhsShape(lhsType.getShape());
+    SmallVector<int64_t> rhsShape(rhsType.getShape());
+    // assuming last 2 input dims are row and col
+    // 0xFF is just a random number > 1, replacing the row and col dims
+    // so that getBroadcastedShape can match, will be removed after
```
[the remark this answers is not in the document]
LongshengDu: added

lib/gc/Transforms/OneDNNGraphToLinalg.cpp
```diff
@@ -270,23 +493,56 @@ struct MatMulOpLowering : public OpRewritePattern<MatMulOp> {
...
     ConversionTarget target(getContext());
     target.addIllegalDialect<onednn_graph::OneDNNGraphDialect>();
-    target.addLegalDialect<BuiltinDialect, arith::ArithDialect,
-                           linalg::LinalgDialect, func::FuncDialect,
-                           tensor::TensorDialect>();
-    // set pattern
+    target.addLegalDialect<
+        // clang-format off
+        BuiltinDialect, 
+        scf::SCFDialect,
```
kurapov-peter, line 519: What are the cases we produce scf during the lowering?

lib/gc/Dialect/OneDNNGraph/OneDNNGraphOps.cpp
```diff
@@ -181,22 +201,37 @@ LogicalResult onednn_graph::MatMulOp::inferReturnTypeComponents(
...
     }
     return success();
   };
+  // get broadcasted batch dims from 2 multi-batch mat shape,
+  auto extractBroadcastBatch = [](ShapedType lhsType, ShapedType rhsType,
+                                  SmallVector<int64_t> &outDims) {
+    SmallVector<int64_t> lhsShape(lhsType.getShape());
+    SmallVector<int64_t> rhsShape(rhsType.getShape());
+    assert(lhsShape.size() >= 2 && rhsShape.size() >= 2);
+    // assuming last 2 input dims are row and col
```
kurapov-peter, line 219: What guarantees it?

kurapov-peter (commented):
One concern I had while reading is the semantics of the shapes folding. I wasn't able to formulate any concrete examples and there might be none, but it still makes sense to give it yet another thought (maybe add complex cases where problems may arise).

lib/gc/Dialect/OneDNNGraph/OneDNNGraphOps.cpp
LongshengDu: In `else if (lRank > 1 && rRank > 1)`, it checks for both input rank >= 2, meaning 2 inputs are all matrix and may have batch dims.

lib/gc/Transforms/OneDNNGraphToLinalg.cpp
```diff
@@ -46,22 +71,39 @@ Value createBroadcastOperand(Location loc, PatternRewriter &rewriter,
...
+    Value newVal = op;
+    if (collapseShape.size() < opShape.size()) {
+      assert(collapseShape.size() + bcastDims.size() == bcastShape.size());
+      auto reassociation =
+          computeReassociationByAnchor(keepDims, opTy.getRank());
+      ShapedType collapseTy =
+          RankedTensorType::get(collapseShape, opTy.getElementType());
+      newVal = rewriter.create<tensor::CollapseShapeOp>(loc, collapseTy, newVal,
+                                                        reassociation);
+    }
```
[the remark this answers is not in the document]
LongshengDu: consider kept dims [a, b,..., c] as anchor, so reassociation = [[...a...], [...b...], ..., [...c...]]  
e.g. for shape [16, 1, 32, 1, 64], rank = 5, kept dims = [0, 2, 4]  
[16, 1, 32, 1, 64] --collapse-> [16, 32, 64]  
reassociation = [[0], [1, 2], [3, 4]]

lib/gc/Dialect/OneDNNGraph/OneDNNGraphOps.cpp
```diff
@@ -59,21 +59,44 @@ SmallVector<int64_t> canonicalizeReduceAxes(ArrayRef<int64_t> axes,
   return ret;
 }
 
-SmallVector<int64_t> getReducedShape(ShapeAdaptor operandShape,
-                                     ArrayRef<int64_t> axes, bool keep_dims) {
-  SmallVector<int64_t> outputShape;
+SmallVector<int64_t> canonicalizeKeepAxes(ArrayRef<int64_t> axes, int64_t rank,
```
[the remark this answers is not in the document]
LongshengDu: added comment

lib/gc/Transforms/OneDNNGraphToLinalg.cpp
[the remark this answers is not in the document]
LongshengDu: added comment

lib/gc/Transforms/OneDNNGraphToLinalg.cpp
LongshengDu: Used to test something, now removed

LongshengDu:
> One concern I had while reading is the semantics of the shapes folding. I wasn't able to formulate any concrete examples and there might be none, but it still makes sense to give it yet another thought (maybe add complex cases where problems may arise).

Can you specify what consider as semantics of the shapes folding? We can look into it.

lib/gc/Transforms/OneDNNGraphToLinalg.cpp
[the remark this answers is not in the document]
kurapov-peter: Thanks. Feels like this can still be better in the following sense. So first, we introduce an "anchor" which is the result of the collapse transformation. Kept dims is an array of indices of the shape. Finally, reassociation would be better off with a clear definition. All these are not obvious and make you read the code to understand. For example, what would happen to `[16, 1, 1, 32]`, rank `4`, kept dims `[0, 3]`? One would expect a `[16, 32]`. What should the reassociation look like?
So, for anchor - do we need another term at all? It overloads the term for fusion as well. Kept dims are clear I think. The reassociation we should also clarify.

lib/gc/Dialect/OneDNNGraph/OneDNNGraphOps.cpp
LongshengDu: Transpose attr only `Controls whether to transpose the last two dimensions`, so batch dims always before last 2 dims according to the [onednn spec](https://spec.oneapi.io/onednn-graph/latest/ops/matrix/MatMul_1.html). If last two dimensions is somehow contain a batch dim, it is definitely wrong.

lib/gc/Transforms/OneDNNGraphToLinalg.cpp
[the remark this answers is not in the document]
LongshengDu: for [16, 1, 1, 32] -> [16, 32], rank = 4, kept dims = [0, 3]
reassociation = [[0], [1, 2, 3]]  
`anchor_dims` is just a internal term/var name for this file, to represent which dims to collapse to/expand from, I don't think it will conflict with the term for fusion.
`reassociation` is not introduced by us, it is the attribute of `tensor.collapse_shape` and `tensor.expand_shape` from [MLIR tensor dialect](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorcollapse_shape-tensorcollapseshapeop)

xurui1995 (approved):
LGTM

yifeizh2 (approved):

lib/gc/Transforms/Pipeline.cpp
```diff
@@ -143,6 +144,8 @@ class GCCPUPipeline : public impl::GCCPUPipelineBase<GCCPUPipeline> {
     auto op = getOperation();
     PassManager pm{op->getContext()};
     populateCPUPipeline(pm);
+    // TODO(longsheng): add a option to
+    // disable threading and enable pm.enableIRPrinting();
```
yifeizh2, line 148: What was this TODO comment for previously ?

LongshengDu: There were no code here previously, I use to add pm.enableIRPrinting() for debug internal passes,  but this function also requires disabling threading, we can figure out later.

kurapov-peter (approved):

Issue #93, opened 2024-05-21
Action clang-tidy is running in a different environment than build


AndreyPavlenko:
It results in the cmake error "OpenMP library could not be found" : https://github.com/intel/graph-compiler/actions/runs/9152095188/job/25159048309#step:10:565

There is also a difference in the cmake command:
https://github.com/intel/graph-compiler/blob/main/.github/workflows/clang-tidy.yml#L75
https://github.com/intel/graph-compiler/blob/main/scripts/compile.sh#L24

I think, it would be reasonable to move clang-tidy into the build script. In this case, it will be running in the same env and reusing the build artifacts, that will make the CI checks a bit faster.

WangJialei-A:
Hi @AndreyPavlenko @ZhennanQin 
`clang-tidy` requires `clang` compiler. 
Using `clang-tidy` and `gcc` together may result in compatibility issues (see [here](https://stackoverflow.com/questions/57082124/how-to-integrate-clang-tidy-to-cmake-and-gcc)). 
So we cannot leverage the compile script or reuse the build artifacts.

For `OpenMP` issue, we need to check whether `libomp-dev` and some related pakages are installed or not in the running image.

WangJialei-A:
Hi, @AndreyPavlenko 
#94 fixed the OpenMP problem. Please reopen this issue if you still have concerns on the `cmake` configuration.

PR #113, opened 2024-05-30
scripts: support local check for developers

WangJialei-A:
support #20

### DIFF
--- a/scripts/full_check.sh
+++ b/scripts/full_check.sh
@@ -48,7 +48,7 @@ done
 # set PR_REF if your pr target is not main
 : ${PR_REF:=main}
 
-PROJECT_ROOT="$(cd "$(dirname "$0")" && pwd)/../"
+PROJECT_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 cd "$PROJECT_ROOT"
 
 MERGE_BASE=$(git merge-base $PR_REF HEAD)
@@ -76,19 +76,16 @@ if [ -n "$CHECK_TIDY" ]; then
 
   cmake ../../ \
     -DCMAKE_BUILD_TYPE=Release \
-    -DMLIR_DIR=${MLIR_DIR} \
+    -DMLIR_DIR="${MLIR_DIR}" \
     -DCMAKE_EXPORT_COMPILE_COMMANDS=True \
     -DCMAKE_C_COMPILER=$(which clang) \
     -DCMAKE_CXX_COMPILER=$(which clang++) \
     -DLLVM_EXTERNAL_LIT=$(which lit)
 
   for f in $(find ./include -name Makefile); do
-    targets=$(make -f $f help |grep IncGen); 
-    if [[ $? -eq 0 ]]; then 
-      for target in $targets; do
-        cd ${f%Makefile} && make ${target#...} && cd -; 
-      done
-    fi ;
+    for target in $(make -f $f help |grep IncGen); do
+      cd ${f%Makefile} && make ${target#...} && cd -; 
+    done
   done
 
   [ -f run-clang-tidy.py ] || wget https://raw.githubusercontent.com/llvm/llvm-project/main/clang-tools-extra/clang-tidy/tool/run-clang-tidy.py -O run-clang-tidy.py
@@ -100,12 +97,12 @@ fi
 if [ -n "$CHECK_FORMAT" ]; then
   echo "start format check..."
   check_tool "clang-format"
-  cd $PROJECT_ROOT
+  cd "$PROJECT_ROOT"
   echo "$CHANGED_FILES" | egrep "*\\.(h|hpp|c|cpp)$" | xargs clang-format --dry-run --Werror -style=file
 fi
 
 if [ -n "$CHECK_LICENSE" ]; then
   echo "start license check..."
-  cd $PROJECT_ROOT
+  cd "$PROJECT_ROOT"
   python3 scripts/license.py --files $(echo $CHANGED_FILES | tr ' ' ',')
 fi
--- a/scripts/full_check.sh
+++ b/scripts/full_check.sh
@@ -24,19 +24,19 @@ check_tool() {
   which $1 &> /dev/null || (echo "$1 not found!" && exit 1)
 }
 
-while [[ $# -gt 0 ]]; do
-  case $1 in
-    --format)
+for arg in "$@"; do
+  case $arg in
+    -f|--format)
       CHECK_FORMAT=1
-      shift
       ;;
-    --tidy)
+    -t|--tidy)
       CHECK_TIDY=1
-      shift
       ;;
-    --license)
+    -l|--license)
       CHECK_LICENSE=1
-      shift
+      ;;
+    -c|--clean)
+      CLEANUP=1
       ;;
     *) # Handle other unknown options
       echo "Unknown option: $1"
@@ -48,9 +48,8 @@ done
 # set PR_REF if your pr target is not main
 : ${PR_REF:=main}
 
-
-PROJECT_ROOT=$(pwd)/$(dirname "${BASH_SOURCE[0]}")/../
-cd $PROJECT_ROOT
+PROJECT_ROOT="$(cd "$(dirname "$0")" && pwd)/../"
+cd "$PROJECT_ROOT"
 
 MERGE_BASE=$(git merge-base $PR_REF HEAD)
 CHANGED_FILES=$(git diff --name-only $MERGE_BASE HEAD)
@@ -70,9 +69,11 @@ if [ -n "$CHECK_TIDY" ]; then
   check_tool "clang-tidy"
   check_tool "lit"
 
-  TIDY_ROOT=${PROJECT_ROOT}/build/tidy
-  mkdir -p ${TIDY_ROOT}
-  cd ${TIDY_ROOT}
+  TIDY_ROOT="${PROJECT_ROOT}/build/tidy"
+  [ -n "$CLEANUP" ] && rm -rf "${TIDY_ROOT}"
+  mkdir -p "${TIDY_ROOT}"
+  cd "${TIDY_ROOT}"
+
   cmake ../../ \
     -DCMAKE_BUILD_TYPE=Release \
     -DMLIR_DIR=${MLIR_DIR} \
--- a/scripts/full_check.sh
+++ b/scripts/full_check.sh
@@ -35,6 +35,11 @@ for arg in "$@"; do
     -l|--license)
       CHECK_LICENSE=1
       ;;
+    -a|--all)
+      CHECK_FORMAT=1
+      CHECK_TIDY=1
+      CHECK_LICENSE=1
+      ;;
     -c|--clean)
       CLEANUP=1
       ;;
--- a/scripts/full_check.sh
+++ b/scripts/full_check.sh
@@ -20,6 +20,10 @@
 # this is for developer local check only.
 # not used by github action
 
+check_tool() {
+  which $1 &> /dev/null || (echo "$1 not found!" && exit 1)
+}
+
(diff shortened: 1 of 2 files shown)


scripts/tidy.sh
```diff
@@ -0,0 +1,60 @@
...
+  -DLLVM_EXTERNAL_LIT=$(which lit)
+
+for f in $(find ./include -name Makefile); do
+  target=$(make -f $f help |grep IncGen); 
+  if [[ $? -eq 0 ]]; then 
+    cd ${f%Makefile} && make ${target#...} && cd -; 
+  fi ;
+done
+
+wget https://raw.githubusercontent.com/llvm/llvm-project/main/clang-tools-extra/clang-tidy/tool/run-clang-tidy.py -O run-clang-tidy.py
```
Menooker, line 57: shall we cache result?

scripts/tidy.sh
```diff
@@ -0,0 +1,60 @@
...
+wget https://raw.githubusercontent.com/llvm/llvm-project/main/mlir/.clang-tidy -O .clang-tidy
```
Menooker, line 58: maybe we can copy this file to our repo? Just like `.clang_format`?

scripts/tidy.sh
```diff
@@ -0,0 +1,60 @@
...
+    exit 1
+fi
+
+# set PR_REF if your pr target is not main
+PR_REF=${PR_REF:-main}
+
+MERGE_BASE=$(git merge-base $PR_REF HEAD)
+CHANGED_FILES=$(git diff --name-only $MERGE_BASE HEAD | paste -sd' ')
+
+mkdir -p build
```
Menooker, line 35: shall we check the current dir is in `scripts`? Or maybe we should automatically cd to `scripts/`

scripts/tidy.sh
```diff
@@ -0,0 +1,60 @@
+#! /bin/bash
```
Menooker, line 1: can we make it a all-in-one checker for clang-format and tidy?

scripts/tidy.sh
WangJialei-A: @Menooker 
I'm not sure it is ok to place code from llvm in our public repository

scripts/tidy.sh
WangJialei-A: good idea

scripts/tidy.sh
Menooker: @kurapov-peter @ZhennanQin What do you think if we copy `.clang-tidy` to our repo?

scripts/tidy.sh
WangJialei-A: @Menooker  OK. I will extend the feature of the script

scripts/tidy.sh
```diff
@@ -0,0 +1,60 @@
...
+fi
+
+# set PR_REF if your pr target is not main
+PR_REF=${PR_REF:-main}
+
+MERGE_BASE=$(git merge-base $PR_REF HEAD)
+CHANGED_FILES=$(git diff --name-only $MERGE_BASE HEAD | paste -sd' ')
+
+mkdir -p build
+cd build
```
AndreyPavlenko, line 36: Maybe use the build dir in the project root? 
```suggestion
BUILD_DIR="$(dirname "$0")/../build/tidy"
mkdir -p "$BUILD_DIR"
cd "$BUILD_DIR"
```

scripts/tidy.sh
AndreyPavlenko: Or do not download if already exists:
```shell
[ -f run-clang-tidy.py ] || wget ...
[ -f .clang-tidy ] || wget ... 
```

scripts/tidy.sh
WangJialei-A: @AndreyPavlenko 
It is the same problem mentioned in https://github.com/intel/graph-compiler/issues/93#issuecomment-2124141173

It is not ensured that developer is using the clang compiler, we cannot reuse that folder.

scripts/full_check.sh
```diff
@@ -0,0 +1,110 @@
...
+      exit 1
+      ;;
+  esac
+done
+
+# set PR_REF if your pr target is not main
+: ${PR_REF:=main}
+
+
+PROJECT_ROOT=$(pwd)/$(dirname "${BASH_SOURCE[0]}")/../
```
AndreyPavlenko, line 52: This will not work if the script is run with the full path, e.g. /a/b/c/full_check.sh you will get /cur/dir//a/b/c/../.
I'd avoid using bash syntax, when you could do the same with POSIX, especially when it's more simple.
```suggestion
PROJECT_ROOT="$(cd "$(dirname "$0")" && pwd)"
```

scripts/full_check.sh
```diff
@@ -0,0 +1,110 @@
...
+cd $PROJECT_ROOT
```
AndreyPavlenko, line 53: The path may contain white spaces.
```suggestion
cd "$PROJECT_ROOT"
```

scripts/full_check.sh
```diff
@@ -0,0 +1,110 @@
...
+    --license)
+      CHECK_LICENSE=1
+      shift
+      ;;
+    *) # Handle other unknown options
+      echo "Unknown option: $1"
+      exit 1
+      ;;
+  esac
+done
```
AndreyPavlenko, line 46:
```suggestion
for arg in "$@"; do
  case $arg in
    -f|--format)
      CHECK_FORMAT=1
      ;;
    -t|--tidy)
      CHECK_TIDY=1
      ;;
    -l|--license)
      CHECK_LICENSE=1
      ;;
    -c|--clean)
      CLEANUP=1
      ;;
    *) # Handle other unknown options
      echo "Unknown option: $1"
      exit 1
      ;;
  esac
done
```

scripts/full_check.sh
```diff
@@ -0,0 +1,110 @@
...
+    exit 1
+  fi
+  check_tool "clang"
+  check_tool "clang++"
+  check_tool "clang-tidy"
+  check_tool "lit"
+
+  TIDY_ROOT=${PROJECT_ROOT}/build/tidy
+  mkdir -p ${TIDY_ROOT}
+  cd ${TIDY_ROOT}
```
AndreyPavlenko, line 75:
```suggestion
  TIDY_ROOT="${PROJECT_ROOT}/build/tidy"
  [ -n "$CLEANUP" ] && rm -rf "${TIDY_ROOT}"
  mkdir -p "${TIDY_ROOT}"
  cd "${TIDY_ROOT}"
```

scripts/tidy.sh
kurapov-peter: > @kurapov-peter @ZhennanQin What do you think if we copy `.clang-tidy` to our repo?

sure, @leshikus please take a look

scripts/full_check.sh
```diff
@@ -0,0 +1,111 @@
...
+    -DCMAKE_BUILD_TYPE=Release \
+    -DMLIR_DIR=${MLIR_DIR} \
+    -DCMAKE_EXPORT_COMPILE_COMMANDS=True \
+    -DCMAKE_C_COMPILER=$(which clang) \
+    -DCMAKE_CXX_COMPILER=$(which clang++) \
+    -DLLVM_EXTERNAL_LIT=$(which lit)
+
+  for f in $(find ./include -name Makefile); do
+    targets=$(make -f $f help |grep IncGen); 
+    if [[ $? -eq 0 ]]; then
```
AndreyPavlenko, line 87:
```suggestion
    if [ $? = 0 ]; then 
```

scripts/full_check.sh
```diff
@@ -0,0 +1,111 @@
+#! /bin/bash
```
AndreyPavlenko, line 1:
```suggestion
#! /bin/bash -e
```

scripts/full_check.sh
```diff
@@ -0,0 +1,111 @@
...
+    exit 1
+  fi
+  check_tool "clang"
+  check_tool "clang++"
+  check_tool "clang-tidy"
+  check_tool "lit"
+
+  TIDY_ROOT="${PROJECT_ROOT}/build/tidy"
+  [ -n "$CLEANUP" ] && rm -rf "${TIDY_ROOT}"
+  mkdir -p "${TIDY_ROOT}"
```
AndreyPavlenko, line 74:
```suggestion
  [ -d "${TIDY_ROOT}" ] || mkdir -p "${TIDY_ROOT}"
```

scripts/full_check.sh
```diff
@@ -0,0 +1,111 @@
...
+  cmake ../../ \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DMLIR_DIR=${MLIR_DIR} \
+    -DCMAKE_EXPORT_COMPILE_COMMANDS=True \
+    -DCMAKE_C_COMPILER=$(which clang) \
+    -DCMAKE_CXX_COMPILER=$(which clang++) \
+    -DLLVM_EXTERNAL_LIT=$(which lit)
+
+  for f in $(find ./include -name Makefile); do
+    targets=$(make -f $f help |grep IncGen);
```
AndreyPavlenko, line 86:
```suggestion
    targets=$(make -f $f help |grep IncGen) ||
```

scripts/full_check.sh
```diff
@@ -0,0 +1,111 @@
...
+      echo "Unknown option: $1"
+      exit 1
+      ;;
+  esac
+done
+
+# set PR_REF if your pr target is not main
+: ${PR_REF:=main}
+
+PROJECT_ROOT="$(cd "$(dirname "$0")" && pwd)/../"
```
AndreyPavlenko, line 51:
```suggestion
PROJECT_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
```

scripts/full_check.sh
leshikus: bash `if` command checks the command exit code by default, the code above is like the check `if mybool == true`

please use either `if cmd` or `cmd && some_command`

scripts/full_check.sh
leshikus: It's usually better to use explicit `set -e` in case someone uses `bash script` to launch the script

scripts/full_check.sh
WangJialei-A: @AndreyPavlenko  
`-p` flag will do nothing if the directory is existing

AndreyPavlenko: If the script is running with -e option, it will fail when the directory exists, because the exit code of mkdir will not be 0.

WangJialei-A: @AndreyPavlenko 


> If the script is running with -e option, it will fail when the directory exists, because the exit code of mkdir will not be 0.

I tried on my local and the exit code is zero if `-p` flag is set for mkdir.

AndreyPavlenko: You are right and it's mentioned in the doc:
```
       -p, --parents
              **no error if existing**, make parent directories as needed, with their file modes unaffected by any -m option.
```

scripts/full_check.sh
```diff
@@ -0,0 +1,108 @@
...
+      CHECK_LICENSE=1
+      ;;
+    -c|--clean)
+      CLEANUP=1
+      ;;
+    *) # Handle other unknown options
+      echo "Unknown option: $1"
+      exit 1
+      ;;
+  esac
```
Yun-Fly, line 45: shall we add another flag to go through all above check, like `-a` or `--all`.

WangJialei-A: @Yun-Fly 
Supported in the following commit.

scripts/full_check.sh
AndreyPavlenko: What about `set -e`? I think, it's very useful for error checking.

AshburnLee (approved):
LGTM

PR #114, opened 2024-05-30
[Dialect] Introduce `microkernel` dialect

huanghaixin008:
Tracking [#12](https://github.com/intel/graph-compiler/issues/12)

This PR introduces `microkernel` dialect for frontend of microkernels, representing four procedures of microkernel execution: 1. dispatch (JIT); 2. prologue (like tile config); 3. execute; 4. epilogue (like tile release); Later passes could directly operate on this dialect for microkernel-related optimization & lowering.

Issue #108, opened 2024-05-29
Build microkernel related pipeline


huanghaixin008:
The pipeline includes:
- Lowering from frontend ops into abstract microkernel layer
- Lowering from microkernel layer into runtime calls
- Optimizations to reduce runtime overhead
- Incorporate oneDNN as runtime backend

PR #116, opened 2024-05-30
[Transform] Introduce `microkernel` dialect optimization passes

huanghaixin008:
Tracking [#108 ](https://github.com/intel/graph-compiler/issues/108)

This PR introduces following pass to optimize microkernel dialect runtime efficiency:

- `EarlyDispatchMicrokernel`: Dispatch microkernel during initialization time to reduce runtime cost
- `InvariantMicrokernelCodeMotion`: Hoist invariant microkernel-related codes to improve performance
- `MergeBranchMicrokernelContext`: Merge and hoist identical microkernel context codes out of branch if possible, enabling further hoist

Depends on PR [#114](https://github.com/intel/graph-compiler/pull/114) and PR [#115](https://github.com/intel/graph-compiler/pull/115)

**NOTE: this PR is not targeting `main` branch now, please DO NOT MERGE.**


lib/gc/Transforms/Microkernel/EarlyDispatchMicrokernel.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+// get or create global kernel handle with initializer, identified by
+// `kernelName`
+static FailureOr<LLVM::GlobalOp>
+getOrCreateGlobalKernelHandle(RewriterBase &rewriter, ModuleOp module,
+                              const std::string &kernelName,
+                              microkernel::BrgemmDispatchOp op) {
+  // Create the global at the entry of the module
+  LLVM::GlobalOp global;
+  if (!(global = module.lookupSymbol<LLVM::GlobalOp>(kernelName))) {
```
Menooker, line 72: Shall we avoid using the return value of C++ assignment, `=`? Maybe we can refactor as
```
LLVM::GlobalOp global = module.lookupSymbol<LLVM::GlobalOp>(kernelName);
if(!global) {
```

test/mlir/test/gc/Dialect/Microkernel/merge-branch-microkernel-context.mlir
```diff
@@ -0,0 +1,278 @@
+// RUN: gc-opt %s -early-dispatch-microkernel -convert-microkernel-to-dnnl-func -cse -merge-branch-microkernel-context -split-input-file | FileCheck %s
```
Menooker, line 1: We'd better to test as less pass as possible in a unit test. Is it possible to check merge-branch optimization only?

lib/gc/Transforms/Microkernel/MicrokernelInvariantCodeMotion.cpp
```diff
@@ -0,0 +1,439 @@
...
+                return false;
+              }
+            }
+            return true;
+          });
+    } else { // brgemmCallType == BrgemmCallType::TILERELEASE
+      expandInvariantScopeWithCond(
+          structInfo, op,
+          [](Operation *op) -> bool {
+            return !llvm::isa<scf::IfOp>(op) &&
```
Menooker, line 367: Suggestion: you can use `llvm::isa<IfOp, IndexSwitchOp>(op)` to check if op is any of these two types.

lib/gc/Transforms/Microkernel/MicrokernelInvariantCodeMotion.cpp
```diff
@@ -0,0 +1,439 @@
...
+}
+
+static BrgemmCallType getBrgemmCallType(Operation *op) {
+  if (!llvm::isa<func::CallOp>(op)) {
+    return BrgemmCallType::INAPPLICABLE;
+  }
+  auto callOp = dyn_cast<func::CallOp>(op);
+  StringAttr callee = callOp.getCalleeAttr().getAttr();
+
+  if (callee == StringAttr::get(op->getContext(), DNNL_BRGEMM_DISPATCH_NAME)) {
```
Menooker, line 49: Do we have the Ops for TILECFG, TILERELEASE, etc.? Will it be easier to match the ops before they are lowered to func calls?

lib/gc/Transforms/Microkernel/MicrokernelInvariantCodeMotion.cpp
```diff
@@ -0,0 +1,439 @@
...
+    return BrgemmCallType::DISPATCH;
+  }
+  if (callee == StringAttr::get(op->getContext(), DNNL_BRGEMM_TILECFG_NAME)) {
```
Menooker, line 52: can we extract the string in the callee instead of comparing it with a new stringattr?

lib/gc/Transforms/Microkernel/MicrokernelInvariantCodeMotion.cpp
huanghaixin008: Currently we don't have Ops equivalent to func calls to TILECFG, TILERELEASE, etc.. Existing microkernel Ops have more abstract semantic and not suitable for this pass; And I think adding direct-mapping-to-func-calls Ops just to simplify this pass seems not very necessary.

lib/gc/Transforms/Microkernel/EarlyDispatchMicrokernel.cpp
huanghaixin008: Refactored, also return early if `global` is valid to avoid large segment of codes wrapped in `if` branch.

test/mlir/test/gc/Dialect/Microkernel/merge-branch-microkernel-context.mlir
huanghaixin008: Fixed, now it run `-merge-branch-microkernel-context` only.

lib/gc/Transforms/Microkernel/MicrokernelInvariantCodeMotion.cpp
huanghaixin008: Fixed.

lib/gc/Transforms/Microkernel/MicrokernelInvariantCodeMotion.cpp
huanghaixin008: Fixed.

PR #118, opened 2024-06-03
Remove use of legacy bufferization passes

Menooker:
Upstream remove the legacy passes https://github.com/llvm/llvm-project/pull/93535/files
We need to switch to one-shot-bufferization solution now.


LongshengDu:
Tested with llma2 mlp graph, successfully lowered to executable.

ZhennanQin (approved):

ciyongch (approved):

Issue #120, opened 2024-06-03
Hoist thread-local allocator within the nested parallel loops


ciyongch:
1) The allocators in parallelized loops are thread-local
2) For nested parallelized loops, one thread might access another thread's thread-local buffer, which will result in unexpected behavior.
3) We need to hoist such allocators for correct result.

lmontigny:
MLP support related, WIP

PR #121, opened 2024-06-03
[Transform] Hoist thread-local allocator within the nested parallel loops

ciyongch:
This PR is to track https://github.com/intel/graph-compiler/issues/120


lib/gc/Transforms/BufferHoist.cpp
```diff
@@ -0,0 +1,375 @@
...
+    return isSequentialLoop(op) &&
+           !dominators->dominates(aliasDominatorBlock, op->getBlock());
+  }
+
+  /// Returns true if the given operation should be considered for hoisting.
+  static bool shouldHoistOpType(Operation *op) { return allowHoisting(op); }
+
+  /// Does not change the internal placement block, as we want to move
+  /// operations out of loops only.
+  void recordMoveToDominator(Block *block) {}
```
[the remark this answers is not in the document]
ciyongch: Yes, true for loop hoist. I keep this structure intact as original one from MLIR.

lib/gc/Transforms/BufferHoist.cpp
```diff
@@ -0,0 +1,375 @@
...
+        MemRefType memrefType = dyn_cast<MemRefType>(allocValue.getType());
+        if (memrefType && memrefType.hasStaticShape()) {
+          Type elementType = memrefType.getElementType();
+          int64_t memrefTypeRank = memrefType.getRank();
+
+          ArrayRef<int64_t> originalShape = memrefType.getShape();
+          // Create a new shape with the desired larger size
+          SmallVector<int64_t, 4> newShape(originalShape.begin(),
+                                           originalShape.end());
+          newShape[0] *= numThreads;
```
ZhennanQin, line 242: Seems you're trying to hoist to the upper level of parallel for?

ciyongch: Good point, it should be hoisted to the top of outer-most parallel for, will fix it.

ZhennanQin: I think we need a discussion about this, that is, which is the right position for thraed-local memory allocation, to leverage thread-local allocator, the hoist will be inside the parallel for. For thread-group memory allocation, I think we want to hoist them to the outer-most parallel for.

ciyongch: As discussed offline, this PR is targeting the 2nd case (thread local allocator within thread-group) only as it's specific to our thread-group usage. While the 1st case mentioned above is already covered by MLIR's `buffer-loop-hoist` pass which is more general.

Menooker:
I have a general question on this pass. After buffer hoisting (including in nested parallel, and normal loops in upstream), the lifetime of the buffer is actually extended, which will make buffer-scheduling produce worse result. Can we consider the lifetime of buffers while hoisting?

The reason is that the semantic of a hoisted buffer is changed

```
for i {
    float a[100]
    use1(a, i)
    use2(a, i)
    float b[100]
    use3(b, i)
    use4(b, i)
}
```

in the above code, every iteration of `i` "creates" a new buffer `a`. So buffers `a` and `b` has non-overlapping lifetime. If we hoist `a` and `b`, the lifetime will change:

```
float a[100]
float b[100]
for i {
    use1(a, i)
    use2(a, i)
    use3(b, i)
    use4(b, i)
}
```

`a` and `b` must have overlapping lifetime, because we don't know if `use1(a, i)` will read the result of the last iteration. If we reuse `a` for `b`, the result may be incorrect.

Menooker:
I can find 3 ways to resolve this lifetime issue

1. Keep hoisting as a independent pass like in this PR, and mark additional info on the hoisted buffers. For example, we can mark on `a` that it is hoisted from `for i`. The obstacle of this solution is how can we mark a "block" as attr of an operation.
2. Adopt the GC V1 way, combine hoisting and scheduling in a single pass. In V1, we first schedule in a parallel-for, and hoist it, and schedule the hoisted buffer with outer parallel-for.
3. Combine hoisting and scheduling in a single pass, with the new upstream-able interfaces. I am redesigning the buffer-scheduling interface for upstream to make it a general "framework" for developers to customize the scheduling behavior. It might be helpful if we can co-design scheduling interface with our special hoisting. In my understanding, our special hoisting will select a non-default allocation-scope for a buffer (by default, buffer-scheduling will choose the nearest surrounding parallel-for. In our case, we will select the outermost parallel-for),  override the size of the buffer (realSize = originSize * numThreads), and override the IR rewrite behavior. However, most of the code can be reused.
I think maybe we can design the interfaces of buffer scheduling, and let hoisting customize it.

ZhennanQin:
> I can find 3 ways to resolve this lifetime issue
> 
> 1. Keep hoisting as a independent pass like in this PR, and mark additional info on the hoisted buffers. For example, we can mark on `a` that it is hoisted from `for i`. The obstacle of this solution is how can we mark a "block" as attr of an operation.
> 2. Adopt the GC V1 way, combine hoisting and scheduling in a single pass. In V1, we first schedule in a parallel-for, and hoist it, and schedule the hoisted buffer with outer parallel-for.
> 3. Combine hoisting and scheduling in a single pass, with the new upstream-able interfaces. I am redesigning the buffer-scheduling interface for upstream to make it a general "framework" for developers to customize the scheduling behavior. It might be helpful if we can co-design scheduling interface with our special hoisting. In my understanding, our special hoisting will select a non-default allocation-scope for a buffer (by default, buffer-scheduling will choose the nearest surrounding parallel-for. In our case, we will select the outermost parallel-for),  override the size of the buffer (realSize = originSize * numThreads), and override the IR rewrite behavior. However, most of the code can be reused.
>    I think maybe we can design the interfaces of buffer scheduling, and let hoisting customize it.

I vote for option 3.

kurapov-peter:
Sounds like yet another case for the cost-modeling'ish approach.

ciyongch:
> I can find 3 ways to resolve this lifetime issue
> 
> 1. Keep hoisting as a independent pass like in this PR, and mark additional info on the hoisted buffers. For example, we can mark on `a` that it is hoisted from `for i`. The obstacle of this solution is how can we mark a "block" as attr of an operation.
> 2. Adopt the GC V1 way, combine hoisting and scheduling in a single pass. In V1, we first schedule in a parallel-for, and hoist it, and schedule the hoisted buffer with outer parallel-for.
> 3. Combine hoisting and scheduling in a single pass, with the new upstream-able interfaces. I am redesigning the buffer-scheduling interface for upstream to make it a general "framework" for developers to customize the scheduling behavior. It might be helpful if we can co-design scheduling interface with our special hoisting. In my understanding, our special hoisting will select a non-default allocation-scope for a buffer (by default, buffer-scheduling will choose the nearest surrounding parallel-for. In our case, we will select the outermost parallel-for),  override the size of the buffer (realSize = originSize * numThreads), and override the IR rewrite behavior. However, most of the code can be reused.
>    I think maybe we can design the interfaces of buffer scheduling, and let hoisting customize it.

Option 3 looks quite promising. 
I would like to emphasis that there're some difference between the buffer hoist in single-thread and multi-thread scenario. For single thread(non-parallel) scenario, the main consideration is for perf perspective, while for multi-thread(parallel) scenario, both functionality (correctness) and perf should be taken into account. That means, for multi-thread scenario, we have to hoist such thread-local allocator out of the parallel loop to avoid any individual thread access another thread's (main thread within any thread-group) local buffer. So I assume buffer-merge pass should mainly focus on the non-parallel scenario?

For the case you mentioned above, doing the buffer-merge firstly then buffer-hoist (the current designed pass order) will resolve the issue? Or maybe you've any other cases for better demonstration?

ciyongch:
> e yet another case for the cost-modeling'ish approach.

I don't think this optimization needs to rely on or contribute to cost-modeling, due to it's always beneficial to hoist buffer in sequential loops (which is already available in upstream), and it's necessary for parallel loops (to avoid access thread-local buffer cross thread), while this usage is currently mainly happened in our project as upstream always use perfect loop nesting for parallel loops.

lib/gc/Transforms/BufferHoist.cpp
ciyongch: It's addressed now.

Menooker:
> And I assume buffer-merge pass should mainly focus on the non-parallel scenario, which means when the pass decides to hoist a buffer it only consider the sequential loops?

You can build parallel-hoisting upon buffer-merge, I suppose. I am updating buffer-merge pass to allow developers to specify 1) where the buffer is allocated after merging (in/out of parallel) 2) the "real" size of the buffer (after parallel hoisting, buffer size expands by `num_threads` times) 3) how to mutate the IR.

Other things can be handled by buffer-merge.

ciyongch:
> > And I assume buffer-merge pass should mainly focus on the non-parallel scenario, which means when the pass decides to hoist a buffer it only consider the sequential loops?
> 
> You can build parallel-hoisting upon buffer-merge, I suppose. I am updating buffer-merge pass to allow developers to specify 1) where the buffer is allocated after merging (in/out of parallel) 2) the "real" size of the buffer (after parallel hoisting, buffer size expands by `num_threads` times) 3) how to mutate the IR.
> 
> Other things can be handled by buffer-merge.

Sure, let's evaluate this function in the new infra, and also the execution order like one-time merge firstly, and then one-time hoisting, or maybe interleaving these two.

lib/gc/Transforms/BufferHoist.cpp
```diff
@@ -0,0 +1,433 @@
...
+  /// Computes the upper bound for the placement block search.
+  Block *computeUpperBound(Block *dominatorBlock, Block *dependencyBlock) {
+    aliasDominatorBlock = dominatorBlock;
+    // If there is a dependency block, we have to use this block as an upper
+    // bound to satisfy all allocation value dependencies.
+    return dependencyBlock ? dependencyBlock : nullptr;
+  }
+
+  /// Returns true if the given operation represents a loop with sequential
+  /// execution semantics and one of the aliases caused the
```
AIYoungcino, line 137: There is a contradiction between the statement "one of the aliases caused the aliasDominatorBlock to be 'above' the block of the given loop operation" and the condition "!dominators->dominates(aliasDominatorBlock, op->getBlock())"？

ciyongch: You're right, the comment here is not correctly reflect the condition below.  Only the `aliasDominatorBlock` is lower (not above) than `op->getBlock()` is the true case.

lib/gc/Transforms/BufferHoist.cpp
```diff
@@ -0,0 +1,433 @@
...
+        // either not exist or is above our parent block. If we cannot move to
+        // our parent operation due to constraints given by the StateT
+        // implementation, break the walk loop. Furthermore, we should not move
+        // allocations out of unknown region-based control-flow operations.
+        if ((!isKnownControlFlowInterface(parentOp) ||
+             !state.isLegalPlacement(parentOp)) &&
+            (hoistLevel == 0))
+          break;
+        if (isa<scf::ForallOp>(parentOp) && hoistLevel > 0)
+          hoistLevel--;
```
AIYoungcino, line 383: "Why is the variable 'hoistLevel' being decreased by 1? This variable is passed by value."

ciyongch: To make sure this thread-local allocator will be hoisted to the outer-most `parallel for` section. For example:
```
scf.forall ...
  scf.forall ...
    scf.forall ...
      %alloc = memref.alloc() ...
```

AIYoungcino: > To make sure this thread-local allocator will be hoisted to the outer-most `parallel for` section. For example:
> 
> ```
> scf.forall ...
>   scf.forall ...
>     scf.forall ...
>       %alloc = memref.alloc() ...
> ```

OK, I meant that the variable 'hoistLevel' is passed by value here, so decrementing it by one does not affect the logic after exiting this function, such as creating a new allocOp or moving the new allocOp in the correct insert point.Therefore, should it actually be passed by reference here?

ciyongch: The value of `hoistLevel` is not related to any `allocOp` but only the `scf.forall` Op, we want the original `hoistLevel` value after exiting this private function. So the value instead of a reference is required here.

ciyongch:
Closed as it's duplicated with a new implementation in https://github.com/intel/graph-compiler/pull/283

PR #126, opened 2024-06-06
Consolidate tests in a single location

kurapov-peter:
Let's keep it clean


ciyongch:
Separating the gtests and lit tests seems more reasonable? And this structure is what upstream MLIR did?

Menooker:
what about a structure like this?

tests
       \
        dnnl
               \
                 dnnl tests
        mlir
              \
                tests
                       \
                         mlir lit tests
                unittest
                        \
                          mlir unittests

kurapov-peter:
> what about a structure like this?

Yeah, sounds good, I've restructured it in a similar way and added a note. Now `mlir` strictly follows upstream structure.

Menooker (approved):

ZhennanQin (approved):

ciyongch (approved):

PR #131, opened 2024-06-13
Transform: support sdpa to flash attention kernel conversion

yifeizh2:


yifeizh2:
As for performance evaluation, there are two issues
- [ ] deep tiled matmul will not be in effect if the parent op of matmul is scf::forall, so we need to either support this scenario or directly invoke the deep tiled matmul in flash attention kernel
- [ ] 5x performance gap

ZhennanQin:
> As for performance evaluation, there are two issues
> 
> * deep tiled matmul will not be in effect if the parent op of matmul is scf::forall, so we need to either support this scenario or directly invoke the deep tiled matmul in flash attention kernel
> * 5x performance gap
> 
> <style> </style>
> SEQ LENGTH / DTYPE	graph compiler v1 (ms)	v2, block 32, brgemm invoked (ms)	Ratio
> 384 / fp32	8.28744	48.996	5.912079
> 768 / fp32	39.1193	190.557	4.871176
> 1536 / fp32	177.446	752.762	4.242203
> 2304 / fp32	389.382	1837.45	4.718888
> 3072 / fp32	682.228	3273.3813	4.798075
> Next steps for performance alignment is
> 
> * [ ]  Compare the precise brgemm config used in both cases (v1 v.s. mlir)
> * [ ]  Perform more detailed performance breakdown

Please try brgemm instead of matmul, which can provide better performance result.

yifeizh2:
> > As for performance evaluation, there are two issues
> > 
> > * deep tiled matmul will not be in effect if the parent op of matmul is scf::forall, so we need to either support this scenario or directly invoke the deep tiled matmul in flash attention kernel
> > * 5x performance gap
> > 
> > <style> </style>
> > SEQ LENGTH / DTYPE	graph compiler v1 (ms)	v2, block 32, brgemm invoked (ms)	Ratio
> > 384 / fp32	8.28744	48.996	5.912079
> > 768 / fp32	39.1193	190.557	4.871176
> > 1536 / fp32	177.446	752.762	4.242203
> > 2304 / fp32	389.382	1837.45	4.718888
> > 3072 / fp32	682.228	3273.3813	4.798075
> > Next steps for performance alignment is
> > 
> > * [ ]   Compare the precise brgemm config used in both cases (v1 v.s. mlir)
> > * [ ]   Perform more detailed performance breakdown
> 
> Please try brgemm instead of matmul, which can provide better performance result.

I dumped the final llvm IR, and verified that the current performance is collected with brgemm invoked. Previously when brgemm was not in effect, the performance is 10x worse. I think I need to do more detailed analysis to find where the performance gap exists exactly.

yifeizh2:
Latest performance:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/yifeizh2/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/yifeizh2/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">


SEQ LENGTH / DTYPE | graph compiler v1 (ms) | v2, block 64, brgemm invoked |  
-- | -- | -- | --
384 / fp32 | 8.28744 | 22.48292223 | 2.712891
768 / fp32 | 39.1193 | 93.392 | 2.387364
1536 / fp32 | 177.446 | 377.7458 | 2.128793
2304 / fp32 | 389.382 | 810.249 | 2.080859
3072 / fp32 | 682.228 | 1514.4907 | 2.219919



</body>

</html>

yifeizh2:
Current observed gap from v1 are the following:

- [ ] No vectorization
- [ ] No fast transpose
- [ ] No post op fusion
- [ ] linalg.exp takes 1/3 of total execution time; needs to convert to optimized version

PR #136, opened 2024-07-01
Update llvm version for pack propagation and linalg memory effects helpers

LongshengDu:
Update llvm version for [[mlir][linalg] Add pattern to bubble-up pack through expand shape op](https://github.com/llvm/llvm-project/pull/93529).  
Update linalg memory effects helpers to latest, following: [[mlir][linalg] Add more precise memory effects to linalg op](https://github.com/llvm/llvm-project/pull/92079).


yifeizh2 (approved):
Compilation failures were resolved. Thanks~

PR #162, opened 2024-07-11
[Transform][BUG] remove static when forward sourceTypeStrs

crazydemo:


Menooker:
Sorry, it is not what I expected.  `some_array_ref = std::array<...>{...}` is always wrong... 

`std::array<...>{...}` is a temp value. It is different from below, which is correct:

```
std::array<...> a{...};
some_array_ref = a;
```

Please note that `a` is a named variable, which is killed at the end of the scope, while `some_array_ref = std::array<...>{...}` kills the array after the end of the statement.

I re-post the comments in the previous PR:

Please always avoid writing code like
```c++
ArrayRef<std::string> x = {"abc"};
somefunc(x);
```

Please note that "abc" has type of `char[4]` which is not `std::string`. The compiler inserts an conversion here, and the code looks like

```c++
ArrayRef<std::string> x = {std::string("abc")};
```

Also note `std::string("abc")` is a temp value and the lifetime ends at the **full evaluation of the statement**, which means it is dead after the `;` at this line. And `x` outlives the temp value and points to a dead pointer.

However, the following code is valid:

```c++
ArrayRef<const char*> x = {"abc"};
ArrayRef<StringRef> y = {"abc"};
```

Below also valid

```
void somefunc(ArrayRef<std::string> x);
....
somefunc({"bf16"});
```

Menooker:
Also I suggest using `std::array<...>{x,y,z}` over `std::array<...>({x,y,z})`. The `()` are redundant, and will trigger ambiguous-error in some old compilers.

Menooker (approved):

PR #167, opened 2024-07-12
Add numpy

leshikus:
Subj

PR #168, opened 2024-07-15
ci: support python binding build

WangJialei-A:
Support python binding build for llvm and gc build
#167 will be paritally covered during install requirements.txt


xurui1995 (approved):
LGTM

AshburnLee (approved):

PR #178, opened 2024-07-19
Fix gc-cpu-runner arith dialect parsing failure

yifeizh2:
Fix gc-cpu-runner parser stage error on bf16 workloads
```
loc("<stdin>":83:53): error: #"arith"<"fastmath<contract>"> : 'none' attribute created with unregistered dialect. If this is intended, please call allowUnregisteredDialects() on the MLIRContext, or use -allow-unregistered-dialect with the MLIR opt tool used
could not parse the input IR
```

### DIFF
--- a/src/gc-cpu-runner/gc-cpu-runner.cpp
+++ b/src/gc-cpu-runner/gc-cpu-runner.cpp
@@ -21,6 +21,7 @@
 #include "mlir/ExecutionEngine/JitRunner.h"
 #include "mlir/ExecutionEngine/OptUtils.h"
 #include "mlir/IR/Dialect.h"
+#include "mlir/InitAllDialects.h"
 #include "mlir/Target/LLVMIR/Dialect/All.h"
 
 #include "llvm/Support/InitLLVM.h"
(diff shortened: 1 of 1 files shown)


yifeizh2:
The above mentioned error is not reproducible on a few environments, so I am not sure whether the fix is a must or not.

src/gc-cpu-runner/gc-cpu-runner.cpp
```diff
@@ -38,6 +39,7 @@ int main(int argc, char **argv) {
   llvm::InitializeNativeTargetAsmParser();
 
   mlir::DialectRegistry registry;
+  mlir::registerAllDialects(registry);
```
Menooker, line 42: I would suggest to find out which dialect we really need, instead of all dialects initialization.

yifeizh2: Only `arith` is needed. I narrowed down the scope.

xurui1995 (approved):
LGTM

crazydemo:
LGTM

crazydemo (approved):

Menooker:
> The above mentioned error is not reproducible on a few environments, so I am not sure whether the fix is a must or not.

CPUs with/without bf16 support may have a difference? I suggest to root-cause this issue a bit more. Anyway, this PR LGTM.

yifeizh2:
> > The above mentioned error is not reproducible on a few environments, so I am not sure whether the fix is a must or not.
> 
> CPUs with/without bf16 support may have a difference? I suggest to root-cause this issue a bit more. Anyway, this PR LGTM.

I checked this issue with @crazydemo last Friday. It happens on 10219, but not 10220, so it is not likely to be related to instruction set.

Issue #146, opened 2024-07-03
const weight packing support


ZhennanQin:
During model inference, model weight is frozen and won't change between iterations. CPU prefers special weight layout to accelerate the execution, then we need to prepack the model weight before model execution. This issue covers below items:

- Analyze how weight pre-packing is done in openvino.
- Provide the RFC about how to do weight pre-packing in MLIR to meet openvino requirement.
- Implement the weight pre-packing pass with current CPU pipeline to support BF16 MLP inference.

ciyongch:
It's beneficial for loading high-precision weight and packing into low precision and cache, while it's not applicable to the INT4 weight in LLM (for example, W4A16 scenario).
We might need to clarify that the original weight's datatype?

ZhennanQin:
> It's beneficial for loading high-precision weight and packing into low precision and cache, while it's not applicable to the INT4 weight in LLM (for example, W4A16 scenario). We might need to clarify that the original weight's datatype?

The first datatype that we want to support is BF16. And I don't think weight packing does not apply to INT4. For W4A16 scenario, we can still use a similar block format like NK8k16n2k for weight packing to improve cache locality when converting INT4 to BF16 before BRGEMM. Any special reason making it not applicable?

ciyongch:
My bad, I was referring to it's not beneficial for the datatype conversion (from INT4 to either FP16 or INT8 higher-precision) step during the entire tensor transformation pipeline.
For re-layout, it's always beneficial from cache locality perspective.

niuxiaog:
In openVINO, the IR of a model consists of its topology and constant values, like weights, in memory.

For each Graph, there is a `GraphContext` attr. A `GraphContext` holds a `WeightsSharing`, which is basically a `std::unordered_map<std::string, MemoryInfo::Ptr>` that stores the memory of cached tensors.

Take `FullyConnectedOp` (`FC` for short) with DNNL primitive as an example. Each `FC` has a `DnnlFCExecutor`, which has an attr of type `ExecutorContext`. The `ExecutorContext` holds an `unordered_map<string, MemoryPtr>` to store the memory of its private cached weights.

In compile stage, the operations (for example, type casting ops) that follow the `ConstantOp` (weights, bias or others) will be executed and the results are cached in the `unordered_map` of the `GraphContext`.

[...]

PR #183, opened 2024-07-24
[RFC] Constant tensors caching pass

niuxiaog:
Related to issue #146 .

### DIFF
--- a/docs/constant_tensors_cache.md
+++ b/docs/constant_tensors_cache.md
@@ -1,33 +1,36 @@
-# Constant weights folding pass
+# Constant tensors folding pass
 
 ## 1 Motivation
 Some tensors of a machine learning model are constant during inference, such as the weights of filters of convolution 
 layers. There are two types of representation in terms of these constant tensors:
 
-- They appear as the arguments of the MLIR module entry function. OneDNN Graph adopts this way. According to the 
-specification of oneDNN Graph, AOT compiling is not supported and the kernel compilation happens with logical tensors 
-instead of real tensors. The literal values of these constant tensors are available at execution time. 
-
-- They appear as literal values within the model, such as `arith.constant` operations in MLIR. 
-**OpenVINO can use this way.** The IR of a OpenVINO model consists of its topology and constant values, like weights, 
+- Type 1: they appear as literal values within the model, such as `arith.constant` operations in MLIR. 
+OpenVINO can use this way. The IR of a OpenVINO model consists of its topology and constant values, like weights, 
 in memory. These values are available in compile time. When transforming OpenVINO IR to MLIR, the constants can be 
 lowered into `arith.constant` operations.
+
+- Type 2: they appear as the arguments of the MLIR module entry function and are marked as constant explicitly. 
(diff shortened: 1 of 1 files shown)


docs/constant_tensors_cache.md
```diff
@@ -0,0 +1,230 @@
...
+```mlir
+module {
+    compute(%feature0: tensor<*xbf16>, %foldedWeight0: tensor<*xbf16>, %foldedWeight1: tensor<*xbf16>, %foldedWeight2: tensor<*xbf16>) 
+            -> %feature3: tensor<*xbf16> {
+        %feature1 = linalg.matmul(%feature0, %foldedWeight0)
+        %feature2 = linalg.matmul(%feature1, %foldedWeight1)
+        %feature3 = linalg.matmul(%feature2, %foldedWeight2)
+        return %feature3
+    }
+    compiletime_fold(%weight1: tensor<*xbf16>) -> %foldedWeight0, %foldedWeight1: tensor<*xbf16>, tensor<*xbf16> {
```
ZhennanQin, line 111: `compiletime_fold` will be a problem because:
1. The kernel binary will contain the whole tensor to be folded, which is too large. If we only want to use `runtime_fold` from it, the binary size is wasted and this is not friendly for kernel cache.
2. For GPU device, we may want to do the folding on CPU. We shouldn't put three functions in the same module to achieve that.

If we want to support direct compile-time folding, I suggest following the direction of section 2.1 to implement it.

niuxiaog: For the compile time available tensors, the integration can choose to:

- lower them to arith.constant, which is not suggested, 
- put them into the arguments list of the module and mark them as `compiletime_const_args`,
- put them into the arguments list of the module and mark them as `runtime_const_args`.

For the first two choices, they will be folded by `compiletime_fold`, and for the third choice, by `runtime_fold`. There will be no large literal tensors in the kernel for the last two choices.

niuxiaog: > We shouldn't put three functions in the same module to achieve that.

I'm not clear if we can generate a new module in the pass pipeline. If so, shall we put the `compiletime_fold` in one module, and `runtime_fold` and `compute` in another module?

ZhennanQin: > I'm not clear if we can generate a new module in the pass pipeline. If so, shall we put the `compiletime_fold` in one module, and `runtime_fold` and `compute` in another module?

I think so. But my current thinking is, we can support `compiletime_fold` in the future when there's a demand for this.

niuxiaog: OK, I will put all folding operations into `runtime_fold`.

lmontigny:
waiting from OV integration, how they handle const weight..?
with TPP -> use it in IR as const. 
to discuss with OV on iteration 4.

lmontigny (approved):
RFC LGTM
Do we have an idea of the performance impact of this pass on some example?

niuxiaog:
> RFC LGTM Do we have an idea of the performance impact of this pass on some example?

Ideally, the execution time of operations on weights of matmuls, such as `tensor.pack` and `linalg.transpose`, will be saved. The ratio of improvement will depend on actual models. We will collect performance data after all is ready.

PR #187, opened 2024-07-25
[Linalgx] Add utils for making and identifying generic packing matmul ops

LongshengDu:
Add utils for making and identifying packing matmul generic ops.  
Supported ops can be easily expanded by adding to `getPackingAttr`. currently support:
- Matmul_4D
- Vnni_Matmul_2D
- Vnni_Matmul_4D
- Vnni_Batch_Reduce_Matmul_3D  

Example:
```cpp
op = makeGenericPackedMatmulOp(builder, loc, linalgx::PackingType::VNNI_MM2D,
                             {tensorA, tensorB}, {tensorC});
isGenericPackedMatmulOp(*op, linalgx::PackingType::VNNI_MM2D); // true
isGenericPackedMatmulOp(*op, linalgx::PackingType::VNNI_MM4D); // false
```


test/mlir/unittests/Linalgx/TestUtils.cpp
```diff
@@ -0,0 +1,235 @@
...
+}
+
+FailureOr<linalg::GenericOp> createMatmulTestBody(OpBuilder &builder,
+                                                  linalgx::VnniOpType opType,
+                                                  Value tensorA, Value tensorB,
+                                                  Value tensorC) {
+  FailureOr<linalg::GenericOp> op = linalgx::makeGenericVnniMatmulOp(
+      builder, builder.getUnknownLoc(), opType, {tensorA, tensorB}, {tensorC});
+  return op;
+}
```
LongshengDu, line 81: See usage here.

lib/gc/Dialect/Linalgx/Utils.cpp
```diff
@@ -0,0 +1,451 @@
...
+  getPackingIteratorTypes(attr.mPacking, utils::IteratorType::parallel);
+  getPackingIteratorTypes(attr.nPacking, utils::IteratorType::parallel);
+  getPackingIteratorTypes(attr.kPacking, utils::IteratorType::reduction);
+  return iteratorTypes;
+}
+
+//****************************************************************************//
+//                     IndexingMaps Utils                                     //
+//****************************************************************************//
+
```
Yun-Fly, line 239: DO we need to align conduct of coding with upstream LLVM? I.e. use `///` commenter instead of `// ***`

lib/gc/Dialect/Linalgx/Utils.cpp
```diff
@@ -0,0 +1,451 @@
...
+                             ArrayRef<SmallVector<AffineExpr> *> exprs) {
+    for (auto &packingMap : mapArray) {
+      auto srcIndex = packingMap.getPackingSrcIndex();
+      auto dstIndex = packingMap.getPackingDstIndex();
+      auto srcDims = packingMap.getPackingSrcDims();
+      auto dstDims = packingMap.getPackingDstDims();
+      auto &dstExprs = *exprs[dstIndex];
+      auto &srcExprs = *exprs[srcIndex];
+      auto compound = getAffineConstantExpr(0, context);
+      for (auto dim : dstDims) {
```
Yun-Fly, line 270: Also community does not prefer `auto` for implicit casting.

test/mlir/unittests/Linalgx/TestUtils.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+  context->appendDialectRegistry(registry);
+  context->loadAllAvailableDialects();
+
+  return OpBuilder(context);
+}
+
+ModuleOp getTestModule(OpBuilder &builder,
+                       std::function<void(OpBuilder &, ValueRange)> createBody,
+                       TypeRange types) {
```
Yun-Fly, line 55: If function `createBody` would not be modified, better to use `function_ref`?

lib/gc/Dialect/Linalgx/Utils.cpp
```diff
@@ -0,0 +1,451 @@
...
+  if (!isa<linalg::GenericOp>(op)) {
+    return false;
+  }
+  // Check for matmul body
+  linalg::GenericOp genericOp = cast<linalg::GenericOp>(op);
+  if (!linalg::detail::isContractionBody(
+          *genericOp.getBlock(), [](Operation *first, Operation *second) {
+            return ((isa<arith::MulFOp>(first) && isa<arith::AddFOp>(second)) ||
+                    (isa<arith::MulIOp>(first) && isa<arith::AddIOp>(second)));
+          })) {
```
Yun-Fly, line 432: Nit: it can be simplified to
```
isContractionBody<
        arith::MulFOp, arith::AddFOp,
        arith::MulIOp, arith::AddIOp>(*genericOp.getBlock())
```

Yun-Fly:
@zhczhong Does this change affect deep tiled matmul?

lib/gc/Dialect/Linalgx/Utils.cpp
```diff
@@ -0,0 +1,451 @@
...
+  // Make the GenericOp
+  return builder.create<linalg::GenericOp>(
+      loc, shapeC, inputs, outputs, indexingMaps, iteratorTypes,
+      [&](OpBuilder &b, Location loc, ValueRange args) {
+        Value result = createMatmulCalc(b, loc, args);
+        b.create<linalg::YieldOp>(loc, result);
+      });
+}
+
+bool isGenericVnniMatmulOp(Operation *op, VnniOpType opType) {
```
huanghaixin008, line 421: For BRGEMM3D, where do we check whether the GenericOp has `batch-reduce` semantic? It seems that a matmul with different semantics on extra dims (other than M, N, K) could also pass the check

lib/gc/Dialect/Linalgx/Utils.cpp
```diff
@@ -0,0 +1,451 @@
...
+      auto secondDims = packingMap.getSecond();
+      for (auto dim : secondDims) {
+        auto size = secondShape.getDimSize(dim);
+        if (size == ShapedType::kDynamic)
+          isDynamic = true;
+        secondSize *= size;
+      }
+      if (isDynamic)
+        return false; // does not support dynamic dims
+      if (firstSize != secondSize)
```
yifeizh2, line 192: Equal size does not necessarily represent the same semantic. Shall we check AffineMap Expr instead?

lib/gc/Dialect/Linalgx/Utils.cpp
```diff
@@ -0,0 +1,451 @@
...
+  }
+  bool batchEqualAC() const {
+    return std::equal(batchA.begin(), batchA.end(), batchC.begin());
+  }
+  SmallVector<int64_t> batchA;
+  SmallVector<int64_t> batchB;
+  SmallVector<int64_t> batchC;
+};
+
+struct PackingMap {
```
yifeizh2, line 66: Please add more explanation for PackingMap for better readability.

lib/gc/Dialect/Linalgx/Utils.cpp
```diff
@@ -0,0 +1,451 @@
...
+  /// SrcDims.size() == 1; DstDims.size() >= 1
+  ArrayRef<int64_t> getPackingSrcDims() const {
+    return getPackingSrcIndex() == 0 ? getFirst() : getSecond();
+  }
+  ArrayRef<int64_t> getPackingDstDims() const {
+    return getPackingDstIndex() == 0 ? getFirst() : getSecond();
+  }
+  /// Index first is 0; Index second is 1
+  unsigned getPackingSrcIndex() const { return getFirst().size() == 1 ? 0 : 1; }
+  unsigned getPackingDstIndex() const { return getFirst().size() == 1 ? 1 : 0; }
```
yifeizh2, line 82: In many cases both first and second can be with `size > 1`, then the result of the utility function might be wrong.

yifeizh2: > In many cases both first and second can be with `size > 1`, then the result of the utility function might be wrong.

Discussed with Longshen, there is no such possibility in matmul

lib/gc/Dialect/Linalgx/Utils.cpp
LongshengDu: This requires `isPairTemplateImpl` and `isContractionBody` in mlir\lib\Dialect\Linalg\IR\LinalgInterfaces.cpp, no exposed outside.

lib/gc/Dialect/Linalgx/Utils.cpp
LongshengDu: Added AffineMap Expr check for any loop order.

lib/gc/Dialect/Linalgx/Utils.cpp
```diff
@@ -0,0 +1,493 @@
...
+  bool isBatchReduce() const {
+    return !isBatchEmpty() && batchEqualAB() && batchC.empty();
+  }
+
+private:
+  bool batchEqualAB() const {
+    return std::equal(batchA.begin(), batchA.end(), batchB.begin());
+  }
+  bool batchEqualAC() const {
+    return std::equal(batchA.begin(), batchA.end(), batchC.begin());
```
zhczhong, line 56: Please check batchA and batchC has the same size otherwise it may throw segfault

LongshengDu: Switch to `llvm::equal` which uses C++14 `std::equal` and check for 2 ranges.

lib/gc/Dialect/Linalgx/Utils.cpp
LongshengDu: Added

lib/gc/Dialect/Linalgx/Utils.cpp
```diff
@@ -0,0 +1,515 @@
...
+  SmallVector<int64_t> second;
+};
+
+/// PackingAttr to represent a matmul packing:
+/// vnni or non-vnni matmul, dim size of weight, batch dims, M,N,K packing map
+struct PackingAttr {
+  bool isVnni = false;
+  int64_t weightDims = 0;
+  BatchDimMap batchDimMap;
+  SmallVector<PackingMap> mPacking;
```
yifeizh2, line 104: Can you also state the detailed relation described by `mPacking`, `nPacking`, and `kPacking`? e.g. `mPacking` is input --> output

Yun-Fly:
I wonder what will return if we call `mlir::linalg::isaContractionOpInterface` on all of these genricOp?

Yun-Fly:
Once this PR merged, there would never appear `linalgX.op`? Is it right?

yifeizh2:
> I wonder what will return if we call `mlir::linalg::isaContractionOpInterface` on all of these genricOp?

I think mm4d will return true, while vnni series ops will return false.

Yun-Fly:
> > I wonder what will return if we call `mlir::linalg::isaContractionOpInterface` on all of these genricOp?
> 
> I think mm4d will return true, while vnni series ops will return false.

Ok, then shall we add an enhanced utility to cover them? I think it looks like a frequent query in our project, at least for fusion part.

LongshengDu:
> Once this PR merged, there would never appear `linalgX.op`? Is it right?

After PR, `linalgx.op` will be considered deprecated and will advice develop to switch. May take a few PRs to migrate all of them.

LongshengDu:
> > > I wonder what will return if we call `mlir::linalg::isaContractionOpInterface` on all of these genricOp?
> > 
> > 
> > I think mm4d will return true, while vnni series ops will return false.
> 
> Ok, then shall we add an enhanced utility to cover them? I think it looks like a frequent query in our project, at least for fusion part.

All vnni matmuls does not comply with `ContractionOpInterface` so maybe we can add a `isMatmul` helper to identify all matmuls including `ContractionOpInterface` and vnni matmuls?

Yun-Fly:
> > > > I wonder what will return if we call `mlir::linalg::isaContractionOpInterface` on all of these genricOp?
> > > 
> > > 
> > > I think mm4d will return true, while vnni series ops will return false.
> > 
> > 
> > Ok, then shall we add an enhanced utility to cover them? I think it looks like a frequent query in our project, at least for fusion part.
> 
> All vnni matmuls do not comply with `ContractionOpInterface` so maybe we can add a `isMatmul` helper to identify all matmuls including `ContractionOpInterface` and vnni matmuls?

Sounds good! That is indeed what we need.

include/gc/Dialect/Linalgx/Utils.h
```diff
@@ -0,0 +1,54 @@
...
+/// @brief identify a generic packed matmul Op based on PackingType
+/// @param op the op
+/// @param opType the PackingType
+/// @return true if op is a generic packed matmul Op
+bool isGenericPackedMatmulOp(Operation *op, PackingType opType);
+
+/// @brief identify a matmul Op based on ContractionOp and PackingType
+/// @param op the op
+/// @return true if op is a matmul Op
+bool isMatmulOp(Operation *op);
```
LongshengDu, line 49: Added `isMatmulOp` interface to check ContractionOp and all PackingType. Maybe there is a more structured way to check for packing matmuls that can replace `NotProjectedPermutations`.

lib/gc/Dialect/Linalgx/Utils.cpp
LongshengDu: Added

lib/gc/Dialect/Linalgx/Utils.cpp
```diff
@@ -0,0 +1,529 @@
...
+      if (firstSize != secondSize)
+        return false;
+    }
+    return true;
+  };
+  bool matchM = matchDims(attr.mPacking, shapeA, shapeC);
+  bool matchN = matchDims(attr.nPacking, shapeB, shapeC);
+  bool matchK = matchDims(attr.kPacking, shapeA, shapeB);
+  bool checkMatch = matchBatch(attr.batchDimMap) && matchM && matchN && matchK;
+  return checkMatch;
```
Yun-Fly, line 214: Better to use `return xx&&yy&&zz` instead, which can return `false` in advance and bypass unnecessary remaining check.

lib/gc/Dialect/Linalgx/Utils.cpp
```diff
@@ -0,0 +1,529 @@
...
+SmallVector<AffineMap> getIndexingMaps(MLIRContext *context, ShapedType shapeA,
+                                       ShapedType shapeB, ShapedType shapeC,
+                                       const PackingAttr &attr) {
+  SmallVector<SmallVector<AffineExpr>> exprsArr;
+  auto dims = getPackingDimsExpr(context, exprsArr, //
+                                 shapeA, shapeB, shapeC, attr);
+  auto mapA = simplifyAffineMap(AffineMap::get(dims, 0, exprsArr[0], context));
+  auto mapB = simplifyAffineMap(AffineMap::get(dims, 0, exprsArr[1], context));
+  auto mapC = simplifyAffineMap(AffineMap::get(dims, 0, exprsArr[2], context));
+  return {mapA, mapB, mapC};
```
Yun-Fly, line 311: There are too many `auto` here. Please follow this [COC](https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable) of LLVM.

Yun-Fly (approved):

yifeizh2 (approved):

PR #200, opened 2024-07-30
Make the project compatible with FetchContent and find_package

AndreyPavlenko:
to simplify integration with thirdparty projects.

 - Use the Gc prefix in the target names.
 - Added a separate shared library for oneDNN integration - GcDnnl.
 - The GcDnnl library and the OneDNNGraphDialect are not built if GC_ENABLE_DNNL=OFF.
 - Added new build options: GC_ENABLE_TEST(renamed), GC_ENABLE_TEST_DNNL, GC_ENABLE_TEST_MLIR, GC_ENABLE_OPT.
 - Renamed the GC_USE_GPU option to GC_ENABLE_GPU to be consistent with other options.
 - Link with the OpenMP library to resolve the "JIT session error: Symbols not found: __kmpc_...".
 - Minor compile.sh script enhancements.


lib/gc/Transforms/OneDNNGraphToLinalg.cpp
```diff
@@ -5,6 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 //===----------------------------------------------------------------------===//
+#ifdef GC_HAS_ONEDNN_DIALECT
```
kurapov-peter, line 8: Can we not include the file in the build instead?

scripts/compile.sh
```diff
@@ -45,6 +49,19 @@ for arg in "$@"; do
     esac
 done
 
+if [ ! -z "$REL_BUILD" ]; then
+    BUILD_TYPE=Release
+elif [ ! -z "$DEV_BUILD" ]; then
+    BUILD_TYPE=Debug
+else
+    BUILD_TYPE=RelWithDebInfo
```
kurapov-peter, line 57: So this is the default now?

scripts/compile.sh
```diff
@@ -73,28 +90,37 @@ build_llvm() {
...
+        -DLLVM_ENABLE_PROJECTS="mlir"\
+        -DLLVM_TARGETS_TO_BUILD="X86" \
+        -DLLVM_BUILD_LLVM_DYLIB=$DYN_LINK \
+        -DLLVM_LINK_LLVM_DYLIB=$DYN_LINK \
+        -DLLVM_INCLUDE_RUNTIMES=OFF \
+        -DLLVM_INCLUDE_EXAMPLES=OFF \
+        -DLLVM_INCLUDE_TESTS=OFF \
+        -DLLVM_INCLUDE_BENCHMARKS=OFF \
+        -DLLVM_INCLUDE_DOCS=OFF \
+        -DLLVM_BUILD_TOOLS=OFF \
```
kurapov-peter, line 112: Tools are nice for development builds.

scripts/compile.sh
```diff
@@ -73,28 +90,37 @@ build_llvm() {
...
+        -DCMAKE_CXX_FLAGS_DEBUG="-g -O0" \
+        -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+        -DLLVM_ENABLE_ASSERTIONS=ON \
+        -DLLVM_ENABLE_PROJECTS="mlir"\
+        -DLLVM_TARGETS_TO_BUILD="X86" \
+        -DLLVM_BUILD_LLVM_DYLIB=$DYN_LINK \
+        -DLLVM_LINK_LLVM_DYLIB=$DYN_LINK \
+        -DLLVM_INCLUDE_RUNTIMES=OFF \
+        -DLLVM_INCLUDE_EXAMPLES=OFF \
+        -DLLVM_INCLUDE_TESTS=OFF \
```
kurapov-peter, line 109: Would this prevent us from running llvm's mlir checks?

test/CMakeLists.txt
```diff
@@ -1,15 +1,8 @@
-if (NOT GC_TEST_ENABLE)
+if (NOT GC_ENABLE_TEST)
     message(STATUS "The tests are not enabled.")
     return()
 endif ()
 
 include(gtest)
-
-# Static graph compiler library to be used in tests
-add_mlir_library(graph_compiler_static STATIC ${GC_LIB_SOURCES})
```
kurapov-peter, line 9: Is there a new target for this?

CMakeLists.txt
```diff
@@ -102,15 +110,62 @@ if(GC_ENABLE_BINDINGS_PYTHON)
...
+        GcCpuRuntime
+        GcGpuPasses
+        GcInterface
+        GcJitWrapper
+        GcPasses
+        GcUtilsIR
+        MLIRCPURuntimeDialect
+        MLIRCPURuntimeTransforms
+        MLIRLinalgx
+        MLIRMicrokernel
```
kurapov-peter, line 135: So these are to cover the static scenario. I think this should eventually be passed through via a global property to be flexible.

CMakeLists.txt
```diff
@@ -24,15 +24,19 @@ set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0")
...
-option(GC_TEST_ENABLE "Build the tests" ON)
-option(GC_USE_GPU "Enable GPU backend" OFF)
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
+
+option(GC_ENABLE_LEGACY ON)
+option(GC_ENABLE_DNNL "Enable the oneDNN library integration" ON)
+option(GC_ENABLE_TEST "Build the tests" ON)
+option(GC_ENABLE_TEST_DNNL "Build the dnnl tests" ${GC_ENABLE_DNNL})
+option(GC_ENABLE_TEST_MLIR "Build the mlir tests" ON)
+option(GC_ENABLE_OPT "Build gc-opt" ON)
```
kurapov-peter, line 34: Can be more generic, i.e. `GC_ENABLE_TOOLS`.

kurapov-peter (approved):
Great patch, thanks! I've put some minor comments and questions. Otherwise looks good. Could you also add a smoke test for exporting functionality? Like an example that uses cmake's `find_package`.

scripts/compile.sh
AndreyPavlenko: Right, the default has not been changed.

test/CMakeLists.txt
AndreyPavlenko: Yes, GcDnnlStatic and it's only enabled if GC_ENABLE_TEST_DNNL.

CMakeLists.txt
AndreyPavlenko: Do you mean so that each module adds its targets to the global property?

kurapov-peter: Yup, this is just a thought.

scripts/compile.sh
AndreyPavlenko: The utils like lit, FileCheck, etc. are enabled

scripts/compile.sh
AndreyPavlenko: Removed this option.

lib/gc/Transforms/OneDNNGraphToLinalg.cpp
kurapov-peter: We can conditionally add it to the global property that forms an appropriate library (be that gcpasses or a separate one, no matter). I had a similar problem in #205

leshikus:
I would suggest splitting the PR into several PRs which match items in the list, it just very hard to follow when we have all the items mixed.

At least I suggest splitting it into two parts: essential and non-essential like renamings.

leshikus (approved):
LGTM I cannot see any style violations, yet as I've said before big changes like this are hard to follow

AndreyPavlenko:
> LGTM I cannot see any style violations, yet as I've said before big changes like this are hard to follow

Sorry for that. Initially it was much smaller...

PR #205, opened 2024-07-31
Rename GC_ENABLE_GPU to GC_ENABLE_IMEX & clean up properties usage

kurapov-peter:
This cleans up the gpu passes target and prepares for the default pipeline.


Menooker:
Overall LGTM. Do we need GC_ENABLE_GPU in the future, when we introduce non-IMEX GPU features?

test/mlir/test/gc/Transforms/GPU/lit.local.cfg
```diff
@@ -1,2 +1,2 @@
-if not config.gc_use_gpu:
+if not config.gc_use_imex:
```
Menooker, line 1: This subdirectory was originally designed for general GPU tests. Shall we rename this directory to IMEX? Or maybe we can add a new subdirectory in test/gc/transforms/GPU/IMEX ?

test/mlir/test/gc/gpu-runner/lit.local.cfg
```diff
@@ -1,2 +1,2 @@
-if not config.gc_use_gpu:
+if not config.gc_use_imex:
```
Menooker, line 1: Similar to above... Do we need to distinguish general GPU tests from IMEX tests?

kurapov-peter:
> Overall LGTM. Do we need GC_ENABLE_GPU in the future, when we introduce non-IMEX GPU features?

I don't yet see a need for that unless it steers some runtime dependencies, unavailable on a cpu-only machine.

test/mlir/test/gc/Transforms/GPU/lit.local.cfg
kurapov-peter: Right, so it should either be less strict and allow some of the tests to run. Another option is to split the tests into different folders. Let's decide once we have them.

kurapov-peter:
This is partially done in #200 and will be completed in #202

Issue #212, opened 2024-08-05
Build failed when GC_ENABLE_DNNL=OFF


xurui1995:
when `GC_ENABLE_DNNL=OFF`, we will not build onednn_graph dialect, but the releated bindings are not removed and test cases are not skipped.

PR #213, opened 2024-08-05
remove bindings and skip tests when GC_ENABLE_DNNL=OFF

xurui1995:
When `GC_ENABLE_DNNL=OFF`,  we still need 
* remove bindings of onednn_graph dialect
* skip test cases using onednn_graph dialect

track: https://github.com/intel/graph-compiler/issues/212


kurapov-peter (approved):
Thanks!

Issue #214, opened 2024-08-06
Lowering linalg softmax op failure


xurui1995:
`linalg.softmax` can not be decomposed and lowered to llvm dialect in gc-cpu-pipeline

example
```
func.func @softmax(%arg0: tensor<2x2x2x2xf32>) -> tensor<2x2x2x2xf32> {
  %0 = tensor.empty() : tensor<2x2x2x2xf32>c
  %1 = linalg.softmax dimension(3)
    ins(%arg0 : tensor<2x2x2x2xf32>) outs(%0 : tensor<2x2x2x2xf32>) -> tensor<2x2x2x2xf32>
  return %1 : tensor<2x2x2x2xf32>
}
```
after `gc-cpu-pipeline`,  `linalg.softmax` still exists
```
module {
  llvm.func @softmax(%arg0: !llvm.ptr, %arg1: !llvm.ptr, %arg2: i64, %arg3: i64, %arg4: i64, %arg5: i64, %arg6: i64, %arg7: i64, %arg8: i64, %arg9: i64, %arg10: i64, %arg11: !llvm.ptr, %arg12: !llvm.ptr, %arg13: i64, %arg14: i64, %arg15: i64, %arg16: i64, %arg17: i64, %arg18: i64, %arg19: i64, %arg20: i64, %arg21: i64) {
    %0 = llvm.mlir.undef : !llvm.struct<(ptr, ptr, i64, array<4 x i64>, array<4 x i64>)>
    %1 = llvm.insertvalue %arg11, %0[0] : !llvm.struct<(ptr, ptr, i64, array<4 x i64>, array<4 x i64>)> 
    %2 = llvm.insertvalue %arg12, %1[1] : !llvm.struct<(ptr, ptr, i64, array<4 x i64>, array<4 x i64>)> 
    %3 = llvm.insertvalue %arg13, %2[2] : !llvm.struct<(ptr, ptr, i64, array<4 x i64>, array<4 x i64>)> 
    %4 = llvm.insertvalue %arg14, %3[3, 0] : !llvm.struct<(ptr, ptr, i64, array<4 x i64>, array<4 x i64>)> 
    %5 = llvm.insertvalue %arg18, %4[4, 0] : !llvm.struct<(ptr, ptr, i64, array<4 x i64>, array<4 x i64>)> 
    %6 = llvm.insertvalue %arg15, %5[3, 1] : !llvm.struct<(ptr, ptr, i64, array<4 x i64>, array<4 x i64>)> 
[...]

PR #215, opened 2024-08-06
Copy DecomposeAggregatedOps pass from tpp-mlir

xurui1995:
Copied from [tpp-mlir project ]( https://github.com/plaidml/tpp-mlir/blob/main/lib/TPP/Transforms/DecomposeAggregatedOps.cpp )
The current MLIR repo does not offer an easy-to-use pass for decomposing `linalg.softmax`, and `tpp-mlir` provide a solution
track: https://github.com/intel/graph-compiler/issues/214


ZhennanQin:
We're waiting the upstream solution from: https://github.com/llvm/llvm-project/pull/97582. @kurapov-peter would you please provide some insights?

xurui1995:
> We're waiting the upstream solution from: [llvm/llvm-project#97582](https://github.com/llvm/llvm-project/pull/97582). @kurapov-peter would you please provide some insights?

Oh, I did not notice this PR on llvm

kurapov-peter:
Unfortunately it won't land in the nearest future. The decision is to create a generic decomposition mechanism (that I'm working on now). I could put a similar implementation into GC if we need it. Do we?

kurapov-peter (approved):

ZhennanQin:
Please rebase.

ZhennanQin (approved):

PR #217, opened 2024-08-06
[Transform][Fusion] Support fusion for K-tiling deep matmul

Yun-Fly:
Fix issue found in llama2-MLP.

A kind reminder: Please DO NOT use rank-altering behavior of `*_slice`. 

```
%extracted_slice_3 = tensor.extract_slice %pack[0, %arg5, 0, 0] [1, 32, 32, 32] [1, 1, 1, 1] : tensor<1x128x32x32xbf16> to tensor<32x32x32xbf16>
%19 = linalgx.batch_reduce_matmul_vnni ins(%extracted_slice_3, %extracted_slice_4 : tensor<32x32x32xbf16>, tensor<32x16x32x2xbf16>) outs(%18 : tensor<32x32xf32>) -> tensor<32x32xf32>
```

Although it is valid and convenient to implicitly reduce rank from `[1, 32, 32, 32]` to `[32, 32, 32]`, fused `packOp` will never yield `[32, 32, 32]`, which will lead to unexpected conflict finally.

Instead, Please explicitly use `tensor.collapse_shape` and `tensor.expand_shape` to alter rank of `*_slice`.

### DIFF
--- a/lib/gc/Transforms/IterativeTilingAndFusion.cpp
+++ b/lib/gc/Transforms/IterativeTilingAndFusion.cpp
@@ -180,6 +180,8 @@ exactTilingOnPackUnPackFilter(RewriterBase &rewriter,
       targetTileSizes = tileSizes;
       DenseMap<int64_t, OpFoldResult> dimAndTileMapping =
           unPackOp.getDimAndTileMapping();
+      if (dimAndTileMapping.empty())
+        return failure();
       targetInnerTileSizes.resize(dimAndTileMapping.size());
       for (const auto &dimAndTile : dimAndTileMapping) {
         targetInnerTileSizes[dimAndTile.first] = dimAndTile.second;
@@ -487,21 +489,10 @@ tileAndFuseConsumerOfOpResult(RewriterBase &rewriter, OpResult result,
 
     if (succeeded(fusedResult)) {
       fusedResultList.push_back(*fusedResult);
-      auto whileProducerOutOfLoopBlock =
-          [&fusedResult](LoopLikeOpInterface loop) -> LogicalResult {
-        Block &body = loop->getRegion(0).front();
-        return failure(fusedResult.value().tiledOps[0]->getBlock() == &body);
-      };
-      SmallVector<LoopLikeOpInterface> outerLoops =
-          scfX::getOuterNestLoopsWhile(
-              (*bestCandidate)->getParentOfType<LoopLikeOpInterface>(),
-              whileProducerOutOfLoopBlock);
-      // g. Manually run cse on region which contains top-level loop of
-      // candidate slice in avoid of conflict with subsequent
-      // `tileAndFuseConsumerOfSlice` get nest loops between next candidate
-      // sliceOp and tiled producer.
-      (void)mlir::simplifyRegions(rewriter,
-                                  {*outerLoops.front()->getParentRegion()});
+      // f. Manually run cse on region which contains original consumer op in
+      // avoid of conflict with subsequent `tileAndFuseConsumerOfSlice` get nest
+      // loops between next candidate sliceOp and tiled producer.
+      (void)mlir::simplifyRegions(rewriter, {*consumer->getParentRegion()});
     }
   }
   if (fusedResultList.empty())
@@ -689,7 +680,7 @@ void iterativeTilingAndFusionUntilExhaustion(
           });
       tiledOps.clear();
       if (changed)
-        (void)mlir::simplifyRegions(rewriter, {f.getRegion()});
+        (void)mlir::simplifyRegions(rewriter, f->getRegions());
     } else {
       // Auto tiling with default tile size if no tiled op found. Follow tiling
       // priority based on OpTy: `Contraction`->`Reduction`->`Elementwise`.
--- a/lib/gc/Transforms/IterativeTilingAndFusion.cpp
+++ b/lib/gc/Transforms/IterativeTilingAndFusion.cpp
@@ -7,6 +7,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "gc/Analysis/TargetDescriptionAnalysis.h"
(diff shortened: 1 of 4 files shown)


lib/gc/Transforms/IterativeTilingAndFusion.cpp
```diff
@@ -543,7 +548,13 @@ LogicalResult iterativelyFuseProducerAndConsumerOfTiledOp(
   return success(numTiledOps > 1);
 }
 
-/// What is self tiled op compared with other fused op?
+/// This is a workaround to deal with LinalgXOp
+static bool isTilableLinalgXOp(Operation *op) {
+  return isa<linalgx::BatchReduceMatmulVnniOp, linalgx::MultiBatchMatmulOp,
```
Yun-Fly, line 553: This is a workaround because `LinalgXOp` is not inherited from `TilingInterface` so far.

lib/gc/Transforms/IterativeTilingAndFusion.cpp
```diff
@@ -553,25 +564,17 @@ LogicalResult iterativelyFuseProducerAndConsumerOfTiledOp(
 ///       yield %5
 ///   }
 /// }
-static LogicalResult isSelfTiledOp(Operation *targetOp) {
-  // 0. check tilable
-  if (!isa<TilingInterface>(targetOp))
+static LogicalResult isTiledOpInLoop(Operation *targetOp) {
+  // 1. check tilable
```
Yun-Fly, line 568: The limitation of single tiled op inside loop has been removed now. @kurapov-peter Please help to take a look :).

ciyongch (commented):
Do we have any new tests to cover the change?

Yun-Fly:
> Do we have any new tests to cover the change?

Sure, added.

lib/gc/Transforms/IterativeTilingAndFusion.cpp
```diff
@@ -656,59 +663,54 @@ void iterativeTilingAndFusionUntilExhaustion(
...
-                                   &changed](Operation *tiledOp) {
-        changed |= succeeded(iterativelyFuseProducerAndConsumerOfTiledOp(
-            rewriter, tiledOp, sliceOptions));
-      });
+      llvm::for_each(
+          tiledOps, [&rewriter, &sliceOptions, &changed](Operation *tiledOp) {
+            changed |= succeeded(iterativelyFuseProducerAndConsumerOfTiledOp(
+                rewriter, tiledOp, sliceOptions));
+          });
+      tiledOps.clear();
```
zhczhong, line 687: The loop will directly return at line 713 even if there are still some untiled ops left because the `tiledOps` has been cleared here.

test/mlir/test/gc/Transforms/iterative-tiling-and-fusion.mlir
```diff
@@ -1,14 +1,6 @@
 // RUN: gc-opt --split-input-file -iterative-tiling-and-fusion %s --cse
 
-module attributes {
```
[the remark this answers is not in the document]
Yun-Fly: Default config is enough for test purpose.

lib/gc/Transforms/IterativeTilingAndFusion.cpp
Yun-Fly: Here `tiledOps.clear();` is in if-branch and used to mark all tiled ops have been visited. But, Line 713 is inside else-branch when `tiledOps` is  empty.  Then, if no new tiled op generated by default tiling, it indeed needs return.

zhczhong: Thanks for the explanation! I didn't notice the return is inside the else block.

lib/gc/Transforms/IterativeTilingAndFusion.cpp
```diff
@@ -68,17 +69,21 @@ getClosestInsertSliceOfResult(OpResult result) {
       sliceOp =
           dyn_cast<OffsetSizeAndStrideOpInterface>(useOfResult.getOwner());
     } else if (auto yieldOp = dyn_cast<scf::YieldOp>(useOfResult.getOwner())) {
-      if (auto loop = dyn_cast<LoopLikeOpInterface>(yieldOp->getParentOp()))
+      if (isa<LoopLikeOpInterface, scf::IfOp>(yieldOp->getParentOp())) {
```
ciyongch, line 72: Will `RegionBranchInterface` better than `scf::IfOp` here?

lib/gc/Transforms/IterativeTilingAndFusion.cpp
```diff
@@ -68,17 +69,21 @@ getClosestInsertSliceOfResult(OpResult result) {
...
         return getClosestInsertSliceOfResult(
-            loop->getResult(useOfResult.getOperandNumber()));
+            yieldOp->getParentOp()->getResult(useOfResult.getOperandNumber()));
+      }
+    } else if (isa<tensor::ExpandShapeOp, tensor::CollapseShapeOp>(
```
ciyongch, line 76: Better to add an `else` clause to capture unexpected cases like this new one?

lib/gc/Transforms/IterativeTilingAndFusion.cpp
Yun-Fly: Thanks for this advice! Changed.

lib/gc/Transforms/IterativeTilingAndFusion.cpp
Yun-Fly: I am afraid not because no branch hit is also expected case in this context.

ciyongch (approved):

zhczhong (approved):

Yun-Fly:
The whole llama2 MLP workload(thanks @yifeizh2 for test input) can successfully run with fusion pass by this patch.

PR #227, opened 2024-08-09
[Transform][Fusion] fix pack/unpack filter

Yun-Fly:
1. Fix Pack/unPack filter logic by `getInnerDimsPos` instead of `getDimAndTileMapping`.
2. When one of multiple consumers failed to fuse, move its operand to the last in use-list in avoid of repeated attempts.
3. Split `linalgOp` and `TensorOp` tiling priority.

### DIFF
--- a/lib/gc/Transforms/IterativeTilingAndFusion.cpp
+++ b/lib/gc/Transforms/IterativeTilingAndFusion.cpp
@@ -643,6 +643,7 @@ static LogicalResult isTiledOpInLoop(Operation *targetOp) {
 
 using OpTileSizeMap = std::unordered_map<std::string, SmallVector<int64_t>>;
 
+/// Default Tiling function only effective for certain `OpTy` operation
 template <typename OpTy>
 static FailureOr<scf::SCFTilingResult>
 defaultTilingOfType(RewriterBase &rewriter, Operation *op,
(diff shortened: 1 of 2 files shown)


lib/gc/Transforms/IterativeTilingAndFusion.cpp
```diff
@@ -157,57 +157,46 @@ exactTilingOnPackUnPackFilter(RewriterBase &rewriter,
 
   SmallVector<OpFoldResult> tileSizes = candidate.getMixedSizes();
   // collect target TileSizes and InnerTileSize to compare
-  SmallVector<OpFoldResult> targetTileSizes, targetInnerTileSizes;
+  SmallVector<OpFoldResult> tileSizesOnInnerDims, innerTiles;
```
yifeizh2, line 160: Please change the comment on line 159 accordingly

Yun-Fly: changed :)

yifeizh2 (approved):

ZhennanQin:
Please create an issue for this PR and some description.

Yun-Fly:
> Please create an issue for this PR and some description.

The issue https://github.com/intel/graph-compiler/issues/232 is created.

lib/gc/Transforms/IterativeTilingAndFusion.cpp
```diff
@@ -683,11 +705,13 @@ void iterativeTilingAndFusionUntilExhaustion(
...
       // Auto tiling with default tile size if no tiled op found. Follow tiling
-      // priority based on OpTy: `Contraction`->`Reduction`->`Elementwise`.
+      // priority based on OpTy:
+      // `ContractionOp`->`ReductionOp`->`LinalgOp`->`TensorOp`.
       SmallVector<DefaultTilingFn> priorityTilingPipeLine = {
           defaultTilingOfType<mlir::linalg::ContractionOpInterface>,
           defaultTilingOfType<mlir::linalg::ReduceOp>,
-          defaultTilingOfType<TilingInterface>};
+          defaultTilingOfType<mlir::linalg::LinalgOp>,
+          defaultTilingOfType<tensor::PackOp, tensor::UnPackOp, tensor::PadOp>};
```
zhczhong, line 714: It's better to preserve the `tilingInterface` as a backup option here in case that other ops will implement `TilingInterface` in the future

Yun-Fly: Ok, good suggestion! Add back now.

zhczhong (approved):

PR #249, opened 2024-08-15
[tests] Re-enable unittesting and partial fix for the existing errors

Menooker:
The unit-testing is accidentally disabled since the last refactoring of the `test` folder. We need to add a dummy folder to hold `lig.cfg.py` for the C++ based unit tests.

This reveals 2 errors in the current repo.
 * memref-to-cpuruntime should be a function pass in the CPUPipeline (this issue is fixed)
 * unit test failure on `TargetDescriptionAnalysis.CPUMissingValue` (**need to fix** in a follow-up PR @zhczhong  )

Also fixed an error which breaks g++9, when we need a workaround on the CXX_FLAGS.


ZhennanQin:
Thank you @Menooker !

CMakeLists.txt
```diff
@@ -29,8 +29,11 @@ list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85827 for details. The issue is
 # fixed in GCC 10.
 if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "10.0")
+  include(CheckCXXCompilerFlag)
   check_cxx_compiler_flag("-Wno-unused-but-set-parameter" CXX_SUPPORTS_WNO_UNUSED_BUT_SET_PARAMETER)
-  append_if(CXX_SUPPORTS_WNO_UNUSED_BUT_SET_PARAMETER "-Wno-unused-but-set-parameter" CMAKE_CXX_FLAGS)
+  if(CXX_SUPPORTS_WNO_UNUSED_BUT_SET_PARAMETER)
```
Menooker, line 34: These lines originally occur after including LLVM cmake, so we don't need to include them by ourselves. A recent refactor swaps the location of these lines with include(LLVM...), so it breaks cmake when g++ < 10.0

test/mlir/unittests/Analysis/TargetDescriptionAnalysisTest.cpp
```diff
@@ -62,7 +62,7 @@ dlti.target_system_spec = #dlti.target_system_spec<
 >} {}
 )mlir";
 
-TEST(TargetDescriptionAnalysis, CPUMissingValue) {
+TEST(DISABLED_TargetDescriptionAnalysis, CPUMissingValue) {
```
Menooker, line 65: We need to re-enable this UT after it is fixed.

crazydemo (approved):

zhczhong (approved):

ciyongch:
So this will fix the escape of unittest from CI/CD, right?

Menooker:
> So this will fix the escape of unittest from CI/CD, right?

Yes. It will!

test/mlir/test/gc/Unit/lit.cfg.py
```diff
@@ -0,0 +1,50 @@
...
+# if none of these are present.
+if "TMP" in os.environ:
+    config.environment["TMP"] = os.environ["TMP"]
+if "TEMP" in os.environ:
+    config.environment["TEMP"] = os.environ["TEMP"]
+
+# Propagate HOME as it can be used to override incorrect homedir in passwd
+# that causes the tests to fail.
+if "HOME" in os.environ:
+    config.environment["HOME"] = os.environ["HOME"]
```
ciyongch, line 34: nit: seems we can also put these into var list below？

test/mlir/unittests/Analysis/TargetDescriptionAnalysisTest.cpp
ciyongch: Please add a TODO/FIXME for future reminder

lib/gc/Transforms/Pipeline.cpp
```diff
@@ -128,7 +128,7 @@ void populateCPURuntimePasses(mlir::OpPassManager &pm) {
   pm.addPass(createForallToParallelLoopPass());
   pm.addPass(createParallelLoopFusionPass());
   pm.addPass(createLoopInvariantCodeMotionPass());
-  pm.addPass(createConvertMemRefToCPURuntime());
+  pm.addNestedPass<func::FuncOp>(createConvertMemRefToCPURuntime());
```
xurui1995, line 131: can we merge this now?  bench tool has an issue for some mlir, and it needs this change.

PR #271, opened 2024-08-21
Support generic op in deep tile matmul and enhance matmul config

zhczhong:
Track https://github.com/intel/graph-compiler/issues/272

### DIFF
--- a/lib/gc/Analysis/MatmulConfigAnalysis.cpp
+++ b/lib/gc/Analysis/MatmulConfigAnalysis.cpp
@@ -37,6 +37,19 @@ static llvm::raw_ostream &operator<<(llvm::raw_ostream &ss,
   return ss;
 }
 
+bool validateConfig(const MatmulConfig &cfg) {
+  if (cfg.MThreads <= 0 || cfg.NThreads <= 0 || cfg.KThreads <= 0 ||
+      cfg.MBlock <= 0 || cfg.NBlock <= 0 || cfg.KBlock <= 0 ||
+      cfg.innerMostMBlock <= 0 || cfg.innerMostNBlock <= 0 ||
+      cfg.innerMostKBlock <= 0)
+    return false;
+  if (cfg.MBlock % cfg.innerMostMBlock != 0 ||
+      cfg.NBlock % cfg.innerMostNBlock != 0 ||
+      cfg.KBlock % cfg.innerMostKBlock != 0)
+    return false;
+  return true;
+}
+
 // generate the candidate for the block size(factor of `num`, pow of 2 which is
 // less than `num`)
 std::vector<uint32_t>
@@ -57,6 +70,8 @@ getCandidate(uint32_t num, uint32_t floor,
     candidates.push_back(candidate);
     candidate *= 2;
   }
+  // In case that no config is valid
+  candidates.push_back(num);
   // remove duplicate candidates
   std::sort(candidates.begin(), candidates.end());
   candidates.erase(std::unique(candidates.begin(), candidates.end()),
@@ -98,10 +113,7 @@ double workloadBalancedCost(linalg::LinalgOp &linalgOp,
                             ArrayRef<uint32_t> shape,
                             const MatmulConfig &config,
                             CPUTargetDescriptionAnalysis &sysDesc) {
-  if (shape.size() < 3) {
-    // Has an invalid shape
-    return 0;
-  }
+  assert(shape.size() >= 3 && "shape.size() should >= 3");
   uint32_t M = shape[0], N = shape[1], K = shape[2];
   uint32_t MTaskNum = llvm::divideCeil(M, config.MBlock);
   uint32_t NTaskNum = llvm::divideCeil(N, config.NBlock);
@@ -122,10 +134,7 @@ double memoryConsumptionOnThreadCost(linalg::LinalgOp &linalgOp,
                                      ArrayRef<uint32_t> shape,
                                      const MatmulConfig &config,
                                      CPUTargetDescriptionAnalysis &sysDesc) {
-  if (shape.size() < 3) {
-    // Has an invalid shape
-    return 0;
-  }
+  assert(shape.size() >= 3 && "shape.size() should >= 3");
   uint32_t M = shape[0], N = shape[1], K = shape[2];
   size_t dtypeSize = DataLayout().getTypeSize(
       ShapeAdaptor(linalgOp.getDpsInputs()[1].getType()).getElementType());
@@ -158,6 +167,36 @@ double computationIntensityOnL2Cache(linalg::LinalgOp &linalgOp,
   if (memoryConsumption * dtypeSize > L2Cache * fullLoadRatio)
     computationIntensity /= outOfCachePenalty;
   return 1 / computationIntensity;
+}
+
+// Bufferization may insert more memref.copy/brgemm cannot verify sucessfully
+// and fall back to linalg lower if the buffer is dynamic
+// Bufferization may insert more memref.copy/brgemm cannot verify sucessfully
+// and fall back to linalg lower if the buffer is dynamic
+double dynamicBufferizationCost(linalg::LinalgOp &linalgOp,
+                                ArrayRef<uint32_t> shape,
+                                const MatmulConfig &config,
+                                CPUTargetDescriptionAnalysis &sysDesc) {
+  assert(validateConfig(config) && "config is invalid");
+  assert(shape.size() >= 3 && "shape.size() should >= 3");
+  uint32_t M = shape[0], N = shape[1];
+  double cost = 0;
+  uint32_t MNumBlockPerThread =
+      llvm::divideCeil(M / config.innerMostMBlock, config.MThreads);
+  uint32_t MNumInnerBlockPerBlock =
+      llvm::divideCeil(config.MBlock, config.innerMostMBlock);
+  uint32_t MCost = MNumBlockPerThread % MNumInnerBlockPerBlock != 0 ||
+                   (M / config.innerMostNBlock % config.MThreads != 0 &&
+                    config.MBlock != config.innerMostMBlock);
+  uint32_t NNumBlockPerThread =
+      llvm::divideCeil(N / config.innerMostNBlock, config.NThreads);
+  uint32_t NNumInnerBlockPerBlock =
+      llvm::divideCeil(config.NBlock, config.innerMostNBlock);
+  uint32_t NCost = NNumBlockPerThread % NNumInnerBlockPerBlock != 0 ||
+                   (N / config.innerMostNBlock % config.NThreads != 0 &&
+                    config.NBlock != config.innerMostNBlock);
+  cost = MCost + NCost;
+  return cost;
 }
 
 using CostModelFn = std::function<double(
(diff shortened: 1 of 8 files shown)


lib/gc/Transforms/Pipeline.cpp
```diff
@@ -95,6 +95,7 @@ void populateBufferizationPasses(mlir::OpPassManager &pm) {
   options.setFunctionBoundaryTypeConversion(
       bufferization::LayoutMapOption::IdentityLayoutMap);
   pm.addPass(bufferization::createOneShotBufferizePass(options));
+  pm.addPass(createCanonicalizerPass());
```
zhczhong, line 98: The `canonicalizerPass` could canonicalize the `scf.for` in memref semantic so that `BufferResultsToOutParams` could remove the last `memref.copy`

huanghaixin008:
The `microkernel` pipeline has not support generic op yet, could we add a temp switch for generating `linalgx` to avoid breaking the pipeline, untill generic support of `microkernel` pipeline is added?

zhczhong:
> The `microkernel` pipeline has not support generic op yet, could we add a temp switch for generating `linalgx` to avoid breaking the pipeline, untill generic support of `microkernel` pipeline is added?

Yes, I could add both supports now. But even on the current pipeline, linalgx op doesn't support bufferization interface and cannot be lowered to memref semantic for microkernel use.

huanghaixin008:
> > The `microkernel` pipeline has not support generic op yet, could we add a temp switch for generating `linalgx` to avoid breaking the pipeline, untill generic support of `microkernel` pipeline is added?
> 
> Yes, I could add both supports now. But even on the current pipeline, linalgx op doesn't support bufferization interface and cannot be lowered to memref semantic for microkernel use.

Now `microkernel` frontend pass has been refactored to be based on linalgx with tensor, so it should be enough

lib/gc/Transforms/DeepTileContractionOp.cpp
```diff
@@ -947,18 +957,19 @@ struct DeepTileMatmul : public OpInterfaceRewritePattern<linalg::LinalgOp> {
 
     // Step 1. Split matmul(bf16xbf16->bf16) to matmul(bf16xbf16->f32) +
     // cast(f32->bf16) if K slicing is needed
-    MatmulConfig cfg =
-        MatmulConfigAnalysis(originOp.getOperation()).getConfig();
-    linalgOp = *linalg::generalizeNamedOp(rewriter, linalgOp);
+    MatmulConfigAnalysis cfgAnalysis =
+        MatmulConfigAnalysis(originOp.getOperation());
+    cfgAnalysis.setAllowUndivisibleInnerBlock(false);
```
yifeizh2, line 962: If we by default not allow indivisible inner block, will we still receive empty valid config for DLRM top/bot MLP workload (where K can be 13 or 479).

test/mlir/test/gc/Transforms/deepTileContractionNamedOp.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: gc-opt --split-input-file --deep-tile-contraction-named-op %s | FileCheck %s
+// RUN: gc-opt --split-input-file %s | FileCheck %s
```
yifeizh2, line 1: Missing `--deep-tile-contraction-op` here?

zhczhong: Thanks for reminder!

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -352,102 +384,110 @@ bool readConfigFromAttrs(MatmulConfig &config, ArrayRef<NamedAttribute> attrs) {
...
-      if (!configCandidates.empty())
-        config = configCandidates[0];
-    }
+      // if there is a given config, skip the cost model
+      if (!hasPredefinedConfig) {
+        LLVM_DEBUG(llvm::dbgs() << "No predefined config\n");
+        // TODO: Could add a weight or priority for cost model
+        SmallVector<std::tuple<CostModelFn, std::string, double>>
+            costModelList = {
+                {dynamicBufferizationCost, "dynamicBufferizationCost", 0.9},
```
ciyongch, line 466: Can you add a simple notes regarding the `weight/priority` used here?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -352,102 +384,110 @@ bool readConfigFromAttrs(MatmulConfig &config, ArrayRef<NamedAttribute> attrs) {
...
-    LLVM_DEBUG(llvm::dbgs()
-               << "Final config\nNumThreads: " << sysDesc.getNumThreads()
-               << ", MatmulConfig: " << config << "\n");
+      LLVM_DEBUG(llvm::dbgs()
+                 << "Final config\nNumThreads: " << sysDesc.getNumThreads()
+                 << ", MatmulConfig: " << config << "\n");
+    }
+    hasConfig = true;
   }
+  return config;
```
ciyongch, line 490: It might be better to validate the `config` before returning? for both calculated and user-provided?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -352,102 +384,110 @@ bool readConfigFromAttrs(MatmulConfig &config, ArrayRef<NamedAttribute> attrs) {
...
-      config.innerMostKBlock = 1;
-      for (auto &&[i, d] : llvm::enumerate(KDimTypeIdx))
-        if (i != 0)
-          config.innerMostKBlock *=
-              linalgOp.getShape(linalgOp.getDpsInputOperand(1))[d];
-      givenInnermostBlock.push_back(config.innerMostKBlock);
-    } else {
-      givenInnermostBlock.push_back(0);
-    }
+      // innermost Block, if the layout is blockied layout, the innermost block
```
ciyongch, line 415: nit: blockied -> blocked

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -160,6 +156,42 @@ double computationIntensityOnL2Cache(linalg::LinalgOp &linalgOp,
...
+  assert(config.innerMostNBlock > 0 &&
+         "innerMostNBlock should be larger than 0");
+  assert(shape.size() >= 3 && "shape.size() should >= 3");
+  uint32_t M = shape[0], N = shape[1];
+  double cost = 0;
+  uint32_t MNumBlock =
+      llvm::divideCeil(M / config.innerMostMBlock, config.MThreads);
+  uint32_t MNumBlockPerThread =
+      llvm::divideCeil(config.MBlock, config.innerMostMBlock);
+  assert(MNumBlockPerThread > 0 && config.MThreads);
```
ciyongch, line 178: As long as `config.MBlock>0`, this is always true, right?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -160,6 +156,42 @@ double computationIntensityOnL2Cache(linalg::LinalgOp &linalgOp,
...
+  uint32_t MCost = (MNumBlock % MNumBlockPerThread != 0 ||
+                    (MNumBlock % config.MThreads != 0 &&
+                     config.MBlock != config.innerMostMBlock));
+  uint32_t NNumBlock =
+      llvm::divideCeil(N / config.innerMostNBlock, config.NThreads);
+  uint32_t NNumBlockPerThread =
+      llvm::divideCeil(config.NBlock, config.innerMostNBlock);
+  assert(NNumBlockPerThread > 0 && config.MThreads);
```
ciyongch, line 186: same as above, BTW, should be `NThreads` here.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -160,6 +156,42 @@ double computationIntensityOnL2Cache(linalg::LinalgOp &linalgOp,
...
+double dynamicBufferizationCost(linalg::LinalgOp &linalgOp,
+                                ArrayRef<uint32_t> shape,
+                                const MatmulConfig &config,
+                                CPUTargetDescriptionAnalysis &sysDesc) {
+  assert(config.MThreads > 0 && "MThreads should be larger than 0");
+  assert(config.NThreads > 0 && "NThreads should be larger than 0");
+  assert(config.innerMostMBlock > 0 &&
+         "innerMostMBlock should be larger than 0");
+  assert(config.innerMostNBlock > 0 &&
+         "innerMostNBlock should be larger than 0");
```
ciyongch, line 170: A common config validator will be good?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -160,6 +156,42 @@ double computationIntensityOnL2Cache(linalg::LinalgOp &linalgOp,
...
+  assert(shape.size() >= 3 && "shape.size() should >= 3");
+  uint32_t M = shape[0], N = shape[1];
+  double cost = 0;
+  uint32_t MNumBlock =
+      llvm::divideCeil(M / config.innerMostMBlock, config.MThreads);
+  uint32_t MNumBlockPerThread =
+      llvm::divideCeil(config.MBlock, config.innerMostMBlock);
+  assert(MNumBlockPerThread > 0 && config.MThreads);
+  uint32_t MCost = (MNumBlock % MNumBlockPerThread != 0 ||
+                    (MNumBlock % config.MThreads != 0 &&
```
ciyongch, line 180: A bit confusion here: `MNumBlock` is more like the M_num_block_per_thread in previous calculation, why do we still need to consider `MNumBlock % config.MThreads` here?

test/mlir/test/gc/Transforms/deepTileContractionNamedOp.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: gc-opt --split-input-file --deep-tile-contraction-named-op %s | FileCheck %s
+// RUN: gc-opt --split-input-file --deep-tile-contraction-op %s | FileCheck %s
```
ciyongch, line 1: Also add a generic version in the test cases?

zhczhong: Added

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: A note is added

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: Thanks for the suggestion! I reuse the `validateConfig` here

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: Changed

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: Yes, Use `validateConfig` to replace this part now

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: Changed! Thanks for the advice

lib/gc/Analysis/MatmulConfigAnalysis.cpp
zhczhong: The code has been modified here. The first term is to check whether `NumBlockPerThread` and `NumInnerBlockPerOuterBlock` is divisible. And the second term is to check whether the `NumBlock` and `NumThreads` is divisible. Both cases will lead to the occurance of dynamic tensor(may insert a lot of `memref.copy` in bufferization and hurt performance)

ciyongch: I see, thanks!

ciyongch (approved):

yifeizh2 (approved):

LongshengDu (approved):

PR #280, opened 2024-08-23
[Transform][Fusion] revert consumer fusion for `tensor.pack`

Yun-Fly:
Since our upstream patch has been merged, revert fusing `tensor.pack` as a consumer.

### DIFF
--- a/include/gc/Transforms/Passes.td
+++ b/include/gc/Transforms/Passes.td
@@ -74,8 +74,11 @@ def IterativeTilingAndFusion : Pass<"iterative-tiling-and-fusion",
     Option<"useCostModel", "use-cost-model", "bool",
            /*default=*/"false",
            "Decide if enable cost model to control iterative fusion.">,
+    Option<"defaultNDTile", "default-nd-tile", "int",
+           /*default=*/"2",
+           "Set default ND-Tile size, such as 2[default] or 3.">,
     ListOption<"defaultTileSize", "default-tile-size", "std::string",
-           "Set default TileSize for the certain type of op, saying `matmul:{32,32}`">,
+           "Set default TileSize for the certain type of op, saying `matmul:{32,32}`.">,
     ];
 }
 def DeepTileContractionOp
--- a/include/gc/Transforms/Passes.td
+++ b/include/gc/Transforms/Passes.td
@@ -60,9 +60,9 @@ def IterativeTilingAndFusion : Pass<"iterative-tiling-and-fusion",
   let description = [{
     The pass tries to fuse any MLIR operation which can be tiled. Moreover, this pass aims to support:
       1. Matmul fusion with element-wise/reduce/broadcast ops.
-      2. Pre-op and post-op fusion.
-      3. Multi-consumer and multi-producer support.
-      4. Multiple level of nest loops and candidates.
+      2. Producer and consumer fusion.
+      3. Arbitrary topology, including residual pattern with multiple consumers .
+      4. Nest loops structure with multiple level candidates.
       5. Flexible option to control the boundary of iterative process.
       6. Default tiling when no op is tiled before fusion.
       7. Cost-model to determine whether to fuse or not.
@@ -74,9 +74,9 @@ def IterativeTilingAndFusion : Pass<"iterative-tiling-and-fusion",
     Option<"useCostModel", "use-cost-model", "bool",
            /*default=*/"false",
            "Decide if enable cost model to control iterative fusion.">,
-    Option<"defaultNDTile", "default-nd-tile", "int",
+    Option<"defaultNDTile", "default-nd-tile", "unsigned",
            /*default=*/"2",
-           "Set default ND-Tile size, such as 2[default] or 3.">,
+           "Set default amount of non-one dimensions in TileSize, such as 1, 2[default, a.k.a. 2D-Tile], etc.">,
     ListOption<"defaultTileSize", "default-tile-size", "std::string",
            "Set default TileSize for the certain type of op, saying `matmul:{32,32}`.">,
     ];
--- a/lib/gc/Transforms/IterativeTilingAndFusion.cpp
+++ b/lib/gc/Transforms/IterativeTilingAndFusion.cpp
@@ -479,8 +479,8 @@ tileAndFuseProducerOfOpOperand(RewriterBase &rewriter, OpOperand &operand,
     return std::nullopt;
 
   // c. Check the producer of root source if is tilable.
-  Operation *producer = realProducer->getDefiningOp<TilingInterface>();
-  if (!producer)
+  Operation *producerOp = realProducer->getDefiningOp<TilingInterface>();
+  if (!producerOp)
     return std::nullopt;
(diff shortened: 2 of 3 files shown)


lib/gc/Transforms/IterativeTilingAndFusion.cpp
```diff
@@ -692,21 +695,31 @@ defaultTilingOfType(RewriterBase &rewriter, Operation *op,
       // All parallel iterator will be tiled by `32` or `16`. If need
       // specified, please set option `defaultTileSize`, like `matmul:{64,64}`.
       if (iterType == utils::IteratorType::parallel) {
-        Range curDomain = iterationDomain[en];
-        std::optional<int64_t> tripCount = mlir::constantTripCount(
-            curDomain.offset, curDomain.size, curDomain.stride);
-        if (tileDims >= 2 && en > 0) {
+// Use 2D-tile by default.
```
ZhennanQin, line 698: It would be better to make this as pass parameter.

Yun-Fly: Good point! Add it to new pass parameter called `default-nd-tile`.

Yun-Fly:
CI seems failed due to issue https://github.com/intel/graph-compiler/issues/281. @WangJialei-A is helping to track this.

test/mlir/test/gc/Transforms/iterative-tiling-and-fusion.mlir
```diff
@@ -339,18 +339,39 @@ module {
...
+
+module {
+  /// CHECK-LABEL: @fuse_pack
+  func.func @fuse_pack(%arg0: tensor<1x32x4096xbf16>, %arg1: tensor<1x32x4096xbf16>) -> tensor<1x1x128x32x32xbf16> {
+    %dest0 = tensor.empty() : tensor<1x32x4096xbf16>
+    /// CHECK: %[[FINAL_RESULT:.*]]:2 = scf.forall (%{{.*}}) = (0, 0) to (1, 4096) step (1, 32)
+    /// CHECK: linalg.add
+    %add = linalg.add ins(%arg0, %arg1 : tensor<1x32x4096xbf16>, tensor<1x32x4096xbf16>) outs(%dest0 : tensor<1x32x4096xbf16>) -> tensor<1x32x4096xbf16>
+    %dest1 = tensor.empty() : tensor<1x1x128x32x32xbf16>
+    /// CHECK: tensor.pack
```
kurapov-peter, line 369: Can it be check-next here (maybe move some ops around)? This would guarantee it's in the same block.

lib/gc/Transforms/IterativeTilingAndFusion.cpp
```diff
@@ -672,18 +680,20 @@ defaultTilingOfType(RewriterBase &rewriter, Operation *op,
...
     defaultTileSize.resize(iteratorTypes.size(), rewriter.getIndexAttr(0));
     // Try tileSize from `32` to `16`.
     SmallVector<int64_t> tsOrder = {32, 16};
-    // Only 2D tile is expected.
-    int tileDims = (isa<mlir::linalg::LinalgOp>(op) && !linalgx::isMatmulOp(op))
-                       ? cast<mlir::linalg::LinalgOp>(op).getNumReductionLoops()
-                       : 0;
+    // Record how many dims have been tiled, including fully tiled, i.e.
+    // tileSize == dimSize.
+    int nonOneTileDims =
```
kurapov-peter, line 693: what does the new name mean?

lib/gc/Transforms/IterativeTilingAndFusion.cpp
```diff
@@ -647,11 +648,18 @@ static LogicalResult isTiledOpInLoop(Operation *targetOp) {
 
 using OpTileSizeMap = std::unordered_map<std::string, SmallVector<int64_t>>;
 
+struct defaultTileConfig {
+  // OpTy-to-TileSize mapping
+  OpTileSizeMap tsMap;
+  // ND-tile size
+  int ndTile;
```
kurapov-peter, line 655:
```suggestion
  int ndTileSize;
```
Does negative tile size make any sense?

include/gc/Transforms/Passes.td
```diff
@@ -74,8 +74,11 @@ def IterativeTilingAndFusion : Pass<"iterative-tiling-and-fusion",
     Option<"useCostModel", "use-cost-model", "bool",
            /*default=*/"false",
            "Decide if enable cost model to control iterative fusion.">,
+    Option<"defaultNDTile", "default-nd-tile", "int",
+           /*default=*/"2",
+           "Set default ND-Tile size, such as 2[default] or 3.">,
```
kurapov-peter, line 79: What's an ND tile?

lib/gc/Transforms/IterativeTilingAndFusion.cpp
Yun-Fly: For a general tensor A, dimensions of which can be [A, B, C, D], would be tiled into tiled_A[E, F, G, H], where E∈[1,A], F∈[1,B],...etc. The `tileSize` controls the size of memory we use to compute or move tensors. Similarly, `nonOneTileDims` is used to represent how many valid tile dimensions(excluding `ts==1`) of `tileSize`. For example, if `tiled_A` has shape of [1,1,C,D], then the `nonOneTileDims` equals to `2`, which maybe utilized by 2D load/store instruction.

include/gc/Transforms/Passes.td
Yun-Fly: `NDTile` decides the maximum of increment variable `nonOneTileDims` as said above. This knob controls the granularity of tiling or fusion.

lib/gc/Transforms/IterativeTilingAndFusion.cpp
Yun-Fly: Thanks for point out! `unsigned` is enough here.

test/mlir/test/gc/Transforms/iterative-tiling-and-fusion.mlir
Yun-Fly: Good idea! Changed.

Yun-Fly:
Current CI failed in `GPU with IMEX` due to outdated LLVM commit there, excluding latest consumer fusion patch for `tensor.pack`.

include/gc/Transforms/Passes.td
kurapov-peter: I think this needs a better name and description. The comment in the knob says it's a tile size. I'd expect something like a [32, 16] as a value for an ND-tile size. A `2` for ND-tile size is not intuitive.

Yun-Fly: `NDTile` is referring to `ndarray` in numpy. As for its description, in avoid of confusion, let me try to rephrase this description to "set default amount of non-one dimensions in `tileSize`"?

kurapov-peter (approved):

ZhennanQin (approved):

PR #283, opened 2024-08-26
[Transform] Hoist thread-local allocator within the nested parallel loops

ciyongch:
This PR is to track https://github.com/intel/graph-compiler/issues/120

This is another implementation of hoisting thread-local allocator base on the new memref-merge design, so it depends on https://github.com/intel/graph-compiler/pull/44


lib/gc/Transforms/MergeAllocTickBased.cpp
```diff
@@ -30,6 +30,69 @@ using namespace special_ticks;
...
+  for (auto v : memType.getShape()) {
+    size *= v;
+  }
+  return size;
+}
+
+static bool needsHoistOutOfParallelLoop(Operation *op) {
+  Operation *parent =
+      op->getParentWithTrait<OpTrait::AutomaticAllocationScope>();
+  if (parent && isa<scf::ForallOp>(parent)) {
```
Menooker, line 47: Please try `isa_and_nonnull` for simplier code

lib/gc/Transforms/MergeAllocTickBased.cpp
```diff
@@ -30,6 +30,69 @@ using namespace special_ticks;
...
+
+  return false;
+}
+
+static bool isForallLoopBoundStatic(Operation *op) {
+  bool isStatic = true;
+  if (auto forallOp = dyn_cast<scf::ForallOp>(op)) {
+    OpBuilder builder{forallOp->getContext()};
+    SmallVector<Value> upperBounds = forallOp.getUpperBound(builder);
+    SmallVector<Value> lowerBounds = forallOp.getLowerBound(builder);
```
Menooker, line 78: It will create constant Op if the bounds are constant (they are originally stored in attr, if bounds are constant). I think using `getMixed***Bound()` should be a better choice. You can use `isa<Value>(....)` to check if it is constant.

lib/gc/Transforms/MergeAllocTickBased.cpp
```diff
@@ -30,6 +30,69 @@ using namespace special_ticks;
...
+
+    isStatic &= llvm::all_of(upperBounds, [](Value &ub) {
+      return (ub.getDefiningOp() &&
+              isa<arith::ConstantIndexOp>(ub.getDefiningOp()));
+    });
+
+    isStatic &= llvm::all_of(lowerBounds, [](Value &lb) {
```
Menooker, line 85: shall we also check the step? I am also not sure if `&=` will short-cut the evaluation of `llvm::all_of(...)` if `isStatic` is false.

test/mlir/test/gc/Transforms/buffer-merge.mlir
```diff
@@ -176,4 +176,185 @@ func.func @different_mem_space() {
...
+    %arg0: memref<2xf32>) {
+    %c4 = arith.constant 4 : index
+    %c16 = arith.constant 16 : index
+    %0 = memref.alloc() : memref<2xf32>
+    scf.forall (%i) in (%c16) {
+      %alloc = memref.alloc() : memref<4xf32> //hoist to outermost
+      scf.forall (%j) in (%c4) {
+        %alloc_0 = memref.alloc() : memref<8xf32>
+        %1 = scf.for %k = %lb to %ub step %step
+          iter_args(%iterBuf = %arg0) -> (memref<2xf32>) {
```
Menooker, line 225: We generally don't set memref in `iter_args`. We will directly write to `arg0` after bufferization. `iter_args` is usually for tensors.

Menooker, line 225: Just to confirm, what are we testing here, to set `iter_args` to `arg0`? Is it to check if the scheduler can skip complex lifetime?

lib/gc/Transforms/MergeAllocTickBased.cpp
```diff
@@ -391,11 +482,63 @@ Value MergeAllocDefaultMutator::buildView(OpBuilder &builder, Block *scope,
...
+                                          mergedAlloc, byteShift, ValueRange{});
+
+  // get the aggregated inductorVar
+  Value inductVar;
+  bool isOuterMostLoop = true;
+  int64_t innerLoopUpperBound = 1;
+  while (parent) {
+    if (auto forallOp = dyn_cast<scf::ForallOp>(parent)) {
+      if (isForallLoopBoundStatic(forallOp)) {
+        SmallVector<Value> upperBounds = forallOp.getUpperBound(builder);
```
Menooker, line 503: suggest to use getStaticUpperBound instead.

lib/gc/Transforms/MergeAllocTickBased.cpp
```diff
@@ -391,11 +482,63 @@ Value MergeAllocDefaultMutator::buildView(OpBuilder &builder, Block *scope,
...
+    Value offsetPerThread =
+        builder.create<arith::MulIOp>(loc, inductVar, origSize);
+    Value byteShiftPerThread =
+        builder.create<arith::AddIOp>(loc, byteShift, offsetPerThread);
+
+    return builder.create<memref::ViewOp>(
+        loc, origAllocOp->getResultTypes().front(), mergedAlloc,
+        byteShiftPerThread, ValueRange{});
+  } else
+    return builder.create<memref::ViewOp>(loc,
```
Menooker, line 539: LLVM coding style suggest to match both `then` and `else` block for `{..}`. Reference: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

See the line `// Use braces for the `if` block to keep it uniform with the `else` block.` :)

lib/gc/Transforms/MergeAllocTickBased.cpp
```diff
@@ -30,6 +30,69 @@ using namespace special_ticks;
 /// and default memory space.
 static bool isMemRefTypeOk(MemRefType type) { return type.hasStaticShape(); }
 
+static inline int64_t getSizeInBytes(MemRefType &memType) {
```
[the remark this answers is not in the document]
ciyongch: My consideration is: this is not a big enhancement in addition to the existing general allocator hoist logic within this "framework", if we're going to unify all the allocator behavior in a separate extension instead of mixing them, I think we can go with this way.

lib/gc/Transforms/MergeAllocTickBased.cpp
ciyongch: sure, we can change to this.

lib/gc/Transforms/MergeAllocTickBased.cpp
ciyongch: Good point, this will create constant Op, I can switch to use the function you mentioned.

lib/gc/Transforms/MergeAllocTickBased.cpp
ciyongch: Yes, I can add the check for the step, and use early return when the expression return false.

test/mlir/test/gc/Transforms/buffer-merge.mlir
ciyongch: Neither, the case is simply to demonstrate the scenario of the mixed usage of `scf.forall` and `scf.for`, the only testing purpose is for those allocators in the case. And BTW, `iter_args` still supports memref in addition to tensor.

lib/gc/Transforms/MergeAllocTickBased.cpp
ciyongch: done.

lib/gc/Transforms/MergeAllocTickBased.cpp
ciyongch: I don't think `forallOp` have `getStaticUpperBound` method.

ciyongch: Seems `getStaticUpperBound()` always return the "std::numeric_limits<int64_t>::min()", so I keep the current impl.

test/mlir/test/gc/Transforms/buffer-merge.mlir
Menooker: We will not generate loops with memref in `iter_args` after bufferization. This feature is for dynamic allocations which will be identified as complex access.

lib/gc/Transforms/MergeAllocTickBased.cpp
Menooker: OK. That may be we are using  constant ops instead of attrs for constant bounds. Another possible way may be to check both MiexedValue and Value being constants.

lib/gc/Transforms/MergeAllocTickBased.cpp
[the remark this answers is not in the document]
Menooker: It will be hard to extract the downstream logic when the upstream PR got merged and if we would like to rebase. :) Just a suggestion. it is up to you anyway :)

lib/gc/Transforms/MergeAllocTickBased.cpp
ciyongch: This doesn't make much difference, both ways shall be fine?

test/mlir/test/gc/Transforms/buffer-merge.mlir
ciyongch: Thanks for your explanation, the complex access is not the main testing purpose for the case here.  And it was originally borrowed from: https://github.com/llvm/llvm-project/blob/main/mlir/test/Dialect/Bufferization/Transforms/buffer-loop-hoisting.mlir#L163-L181.

Menooker (approved):

lib/gc/Transforms/MergeAllocTickBased.cpp
```diff
@@ -180,28 +237,56 @@ bool TickCollecter::isMergeableAlloc(TickCollecterStates *s, Operation *op,
...
+        op->getParentWithTrait<OpTrait::AutomaticAllocationScope>();
+    while (auto forallOp = dyn_cast<scf::ForallOp>(parent)) {
+      if (!isForallLoopBoundStatic(forallOp))
+        break;
+
+      OpBuilder builder{forallOp->getContext()};
+      SmallVector<Value> ubs = forallOp.getUpperBound(builder);
+      if (std::optional<int64_t> ubs0_int = getConstantIntValue(ubs[0])) {
+        int64_t innerLoopUpperBound = ubs0_int.value();
+        numThreads *= innerLoopUpperBound;
```
Yun-Fly, line 278: Do we assume that the lower bound must be `0` and step is `1`? Is it general?

ciyongch: Yes, this feature is not a general case, only targeting the downstream usage.

Yun-Fly: All right, then shall we add more assumption check in `isForallLoopBoundStatic`?

ciyongch: Done.

Yun-Fly (approved):

lib/gc/Transforms/MergeAllocTickBased.cpp
```diff
@@ -30,6 +30,62 @@ using namespace special_ticks;
...
+  if (!forallOp)
+    return false;
+
+  for (auto [lb, ub, step] :
+       llvm::zip(forallOp.getMixedLowerBound(), forallOp.getMixedUpperBound(),
+                 forallOp.getMixedStep())) {
+    std::optional<int64_t> ubConst = getConstantIntValue(ub);
+    return ubConst.has_value() && isConstantIntValue(lb, 0) &&
+           isConstantIntValue(step, 1);
+  }
```
Yun-Fly, line 84: BTW, is `return` here suitable? It seems only check the first bound of `scf.forall`.

lib/gc/Transforms/MergeAllocTickBased.cpp
```diff
@@ -30,6 +30,62 @@ using namespace special_ticks;
...
+static bool isForallLoopBoundStatic(Operation *op) {
+  auto forallOp = dyn_cast<scf::ForallOp>(op);
+  if (!forallOp)
+    return false;
+
+  for (auto [lb, ub, step] :
+       llvm::zip(forallOp.getMixedLowerBound(), forallOp.getMixedUpperBound(),
+                 forallOp.getMixedStep())) {
+    std::optional<int64_t> ubConst = getConstantIntValue(ub);
+    return ubConst.has_value() && isConstantIntValue(lb, 0) &&
```
zhczhong, line 82: Why not support the case with `lb != 0` or `step != 1`?  We may have loop like
```
scf.forall (%arg7) = (0) to (512) step (32)
```

lib/gc/Transforms/MergeAllocTickBased.cpp
```diff
@@ -180,28 +236,56 @@ bool TickCollecter::isMergeableAlloc(TickCollecterStates *s, Operation *op,
...
-  int64_t size = refType.getElementTypeBitWidth() / 8;
-  // treat bool (i1) as 1 byte. It may not be true for all targets, but we at
-  // least have a large enough size for i1
-  size = (size != 0) ? size : 1;
-  for (auto v : refType.getShape()) {
-    size *= v;
+
+  // Get the total number of threads from the outermost to the current level of
+  // the parallel loop that the allocation located in.
+  int64_t numThreads = 1;
```
zhczhong, line 265: The calculation of `numThreads` could directly use the upstream util `constantTripCount`.

lib/gc/Transforms/MergeAllocTickBased.cpp
ciyongch: updated.

lib/gc/Transforms/MergeAllocTickBased.cpp
ciyongch: This is not a hard limitation, it's now updated to support this case.

lib/gc/Transforms/MergeAllocTickBased.cpp
ciyongch: Thanks for the tips, changed to use the util function.

lib/gc/Transforms/MergeAllocTickBased.cpp
```diff
@@ -180,28 +239,58 @@ bool TickCollecter::isMergeableAlloc(TickCollecterStates *s, Operation *op,
...
+  if (needsHoistOutOfParallelLoop(op)) {
+    Operation *parent =
+        op->getParentWithTrait<OpTrait::AutomaticAllocationScope>();
+    while (auto forallOp = dyn_cast<scf::ForallOp>(parent)) {
+      if (!isForallLoopBoundStatic(forallOp))
+        break;
+
+      OpBuilder builder{forallOp->getContext()};
+      std::optional<int64_t> numIterations = constantTripCount(
+          forallOp.getLowerBound(builder)[0],
```
zhczhong, line 278: Better to add a check that there is only one induction variable in the forall loop

ciyongch: We can also support multiple induction vars if it's possible in the template design, do we have that case?

zhczhong (approved):

Issue #279, opened 2024-08-23
Corrupt stack detected at runtime


WangJialei-A:
Please try to reproduce the issue on PR #161 
This issue may share the same root cause with #278, but not the same phenomena

reproduce command 
```bash
python -m benchgc --verbose 1 --driver mlir --case /path/to/mlir
python: /home/jovyan/graph-compiler/lib/gc/ExecutionEngine/CPURuntime/MemoryPool.cpp:213: void (anonymous namespace)::FILOMemoryPool::dealloc(void *): Assertion `chunk->canary == MemoryChunk::magicCheckNum && "Corrupt stack detected"' failed.
```

Please save the following module as an mlir file
```mlir
module {
  func.func @entry(%arg0: tensor<1x32x4096xbf16>, %arg1: tensor<4096x4096xbf16>, %arg2: tensor<1x32x4096xbf16>, %arg3: tensor<1xf32>, %arg4: tensor<4096xbf16>, %arg5: tensor<4096x11008xbf16>, %arg6: tensor<4096x11008xbf16>, %arg7: tensor<11008x4096xbf16>, %arg8: tensor<1xf32>, %arg9: tensor<4096xbf16>) -> tensor<1x32x4096xbf16> attributes {llvm.emit_c_interface} {
    %collapsed = tensor.collapse_shape %arg0 [[0, 1], [2]] : tensor<1x32x4096xbf16> into tensor<32x4096xbf16>
    %cst = arith.constant 0.000000e+00 : bf16
    %0 = tensor.empty() : tensor<32x4096xbf16>
    %1 = linalg.fill ins(%cst : bf16) outs(%0 : tensor<32x4096xbf16>) -> tensor<32x4096xbf16>
    %2 = linalg.matmul ins(%collapsed, %arg1 : tensor<32x4096xbf16>, tensor<4096x4096xbf16>) outs(%1 : tensor<32x4096xbf16>) -> tensor<32x4096xbf16>
    %expanded = tensor.expand_shape %2 [[0, 1], [2]] output_shape [1, 32, 4096] : tensor<32x4096xbf16> into tensor<1x32x4096xbf16>
    %3 = tensor.empty() : tensor<1x32x4096xbf16>
    %4 = linalg.add ins(%arg2, %expanded : tensor<1x32x4096xbf16>, tensor<1x32x4096xbf16>) outs(%3 : tensor<1x32x4096xbf16>) -> tensor<1x32x4096xbf16>
[...]

Issue #278, opened 2024-08-23
bf16 MLP failed with memory pool enabled


yifeizh2:
DLRM bot workload (`LD_PRELOAD=/opt/miniforge3/lib/libiomp5.so:/home/yifei/ipex_env/gperftools-2.7.90/.libs/libtcmalloc.so numactl -C 0-55 -m 0 python3 ./tools/main.py --driver=mlp --batch_size=128 --hidden_size_list=13x512x256x128 --has_bias=512x256x128 --act_type=relu --dtype=bf16`) failed with the following error message

```
python3: ../lib/gc/ExecutionEngine/CPURuntime/MemoryPool.cpp:215: void {anonymous}::FILOMemoryPool::dealloc(void*): Assertion `current->allocated > chunk->size' failed.
bench_mlp.sh: line 9: 1393883 Aborted                 (core dumped) LD_PRELOAD=/opt/miniforge3/lib/libiomp5.so:/home/yifei/ipex_env/gperftools-2.7.90/.libs/libtcmalloc.so numactl -C 0-55 -m 0 python3 ./tools/main.py --driver=mlp --batch_size=128 --hidden_size_list=13x512x256x128 --has_bias=512x256x128 --act_type=relu --dtype=bf16 -p
```

More analysis shall be performed. Current observation is that running any one of the layers alone will not trigger the above-mentioned issue.

Issue #284, opened 2024-08-26
Correctness check fail


WangJialei-A:
Please try module to reproduce the problem
position with the form (0, 8k -1) always get inf answer 
```mlir
module {
  func.func @entry(%arg0: tensor<16x32x32xbf16>, %arg1: tensor<16x32x32xbf16>) -> tensor<16x32xf32> attributes {llvm.emit_c_interface} {
    %cst_0 = arith.constant dense<2.000000e+00> : tensor<16x32x32xf32>
    %dest0 = tensor.empty() : tensor<16x32x32xbf16>
    %0 = linalg.add ins(%arg0, %arg1 : tensor<16x32x32xbf16>, tensor<16x32x32xbf16>) outs(%dest0 : tensor<16x32x32xbf16>) -> tensor<16x32x32xbf16>

    %5 = tensor.empty() : tensor<16x32x32xf32>
    %6 = linalg.copy ins(%0 : tensor<16x32x32xbf16>) outs(%5 : tensor<16x32x32xf32>) -> tensor<16x32x32xf32>

    %7 = tensor.empty() : tensor<16x32x32xf32>
    %8 = linalg.powf ins(%6, %cst_0 : tensor<16x32x32xf32>, tensor<16x32x32xf32>) outs(%7 : tensor<16x32x32xf32>) -> tensor<16x32x32xf32>
    %dest1 = tensor.empty() : tensor<16x32xf32>
    %1 = linalg.reduce { arith.addf } ins(%8 : tensor<16x32x32xf32>) outs(%dest1 : tensor<16x32xf32>) dimensions = [2]
    return %1 : tensor<16x32xf32>
  }
}
```
benchgc result
```
              (0, 7): ref:   30.5999756 res:          inf abs_diff:          inf rel_diff:          inf
              (0, 8): ref:   46.1308594 res:  554.3241577 abs_diff:  508.1932983 rel_diff:   11.0163412
              (0, 9): ref:   54.8237228 res:  224.7723694 abs_diff:  169.9486389 rel_diff:    3.0999105
             (0, 10): ref:   62.6171188 res:  266.4804688 abs_diff:  203.8633423 rel_diff:    3.2557125
             (0, 11): ref:   49.8187370 res:  331.9580688 abs_diff:  282.1393433 rel_diff:    5.6633177
             (0, 12): ref:   91.7294464 res:  290.1029053 abs_diff:  198.3734589 rel_diff:    2.1625929
[...]

PR #287, opened 2024-08-27
[Transform] Only use gc runtime allocator for stack-like alloca ops

crazydemo:
use `createPromoteBuffersToStackPass` to convert all legal stack-like allocations into alloca, and only do gc runtime allocator transformation on `allocaOp`, the corresponding dealloc is inserted.

Fix [Issue#279](https://github.com/intel/graph-compiler/issues/279), [Issue#278](https://github.com/intel/graph-compiler/issues/278)


lib/gc/Transforms/MemRefToCPURuntime.cpp
```diff
@@ -96,6 +97,26 @@ struct ConvertMemRefToCPURuntime
               }
             }
           }
+        } else if (isa<memref::AllocOp>(op)) {
+          allocStack.push_back(op);
+        } else if (isa<memref::DeallocOp>(op)) {
+          // fallback alloc / dealloc not in FILO fashion
+          Value deallocMemref = op->getOperands().front();
+          auto aliases = analysis.resolveReverse(deallocMemref);
```
ciyongch, line 105: Do we really need the aliases for analyzing deallocMemref vs AllocMemref? They'll be exactly same, right?

ciyongch (commented):
It will be good to add a simple case to cover this transform fallback instead of throwing error?

lib/gc/Transforms/MemRefToCPURuntime.cpp
crazydemo: According to `createBufferDeallocationPass`, `deallocMemref` will be the same as `AllocMemref`. But just in case we will have an IR as below, and directly call the `memref-to-cpuruntime pass`. Do you think we can omit this possibility, and do not use alias analysis?
```
func.func @alloc_dealloc_view() {
  // CHECK-LABEL: func @alloc_dealloc_view()
  // CHECK: %[[m0:.*]] = cpuruntime.alloc() : memref<128xi8>
  %c0 = arith.constant 0: index
  %f0 = arith.constant 0.0: f32
  %alloc = memref.alloc() : memref<128xi8>
  %view = memref.view %alloc[%c0][] : memref<128xi8> to memref<32xf32>
  %subview = memref.subview %view[0][16][1] : memref<32xf32> to memref<16xf32>
  // CHECK: cpuruntime.dealloc
  memref.dealloc %subview : memref<16xf32>
  return
}
```

Menooker: Why do we have such case when it is not stack-like?

crazydemo: This case is from [Issue#279](https://github.com/intel/graph-compiler/issues/279). After `bufferDeallocation pass`, we will get the following IR, whose `alloc / dealloc` are in `FIFO` fashion. This case should be resolved by [PR#283](https://github.com/intel/graph-compiler/pull/283), with `MergeAlloc pass` enabled, thus the small temp buffers will be merged into a larger buffer. 
It will be better that the stack-like alloc / dealloc to be checked, in case we still have some non-stack like allocations in some complex scenarios.
```
scf.parallel (%arg11) = (%c0_3) to (%c32_4) step (%c16_5) {
      %alloc_18 = memref.alloc() {alignment = 64 : i64} : memref<1x16x4096xbf16>
      %subview = memref.subview %arg2[0, %arg11, 0] [1, 16, 4096] [1, 1, 1] : memref<1x32x4096xbf16> to memref<1x16x4096xbf16, strided<[131072, 4096, 1], offset: ?>>
      %subview_19 = memref.subview %expand_shape[0, %arg11, 0] [1, 16, 4096] [1, 1, 1] : memref<1x32x4096xbf16> to memref<1x16x4096xbf16, strided<[131072, 4096, 1], offset: ?>>
      scf.parallel (%arg12, %arg13, %arg14) = (%c0, %c0, %c0) to (%c1, %c16, %c4096) step (%c1, %c1, %c1) {
        %3 = memref.load %subview[%arg12, %arg13, %arg14] : memref<1x16x4096xbf16, strided<[131072, 4096, 1], offset: ?>>
        %4 = memref.load %subview_19[%arg12, %arg13, %arg14] : memref<1x16x4096xbf16, strided<[131072, 4096, 1], offset: ?>>
        %5 = arith.extf %4 fastmath<contract> : bf16 to f32
        %6 = arith.extf %3 fastmath<contract> : bf16 to f32
        %7 = arith.addf %6, %5 : f32
        %8 = arith.truncf %7 fastmath<contract> : f32 to bf16
        memref.store %8, %alloc_18[%arg12, %arg13, %arg14] : memref<1x16x4096xbf16>
        scf.reduce 
      }
      %alloc_20 = memref.alloc() {alignment = 64 : i64} : memref<1x16x4096xf32>
      scf.parallel (%arg12, %arg13, %arg14) = (%c0, %c0, %c0) to (%c1, %c16, %c4096) step (%c1, %c1, %c1) {
        %3 = memref.load %alloc_18[%arg12, %arg13, %arg14] : memref<1x16x4096xbf16>
        %4 = arith.extf %3 : bf16 to f32
        memref.store %4, %alloc_20[%arg12, %arg13, %arg14] : memref<1x16x4096xf32>
        scf.reduce 
      }
      memref.dealloc %alloc_18 : memref<1x16x4096xbf16>
      %alloc_21 = memref.alloc() {alignment = 64 : i64} : memref<1x16x4096xf32>
      scf.parallel (%arg12, %arg13, %arg14) = (%c0, %c0, %c0) to (%c1, %c16, %c4096) step (%c1, %c1, %c1) {
        %3 = memref.load %alloc_20[%arg12, %arg13, %arg14] : memref<1x16x4096xf32>
        %4 = memref.load %0[%arg12, %arg13, %arg14] : memref<1x16x4096xf32>
        %5 = math.powf %3, %4 : f32
        memref.store %5, %alloc_21[%arg12, %arg13, %arg14] : memref<1x16x4096xf32>
        scf.reduce 
      }
      memref.dealloc %alloc_20 : memref<1x16x4096xf32>
      %alloc_22 = memref.alloc() {alignment = 64 : i64} : memref<1x16xf32>
      scf.parallel (%arg12, %arg13) = (%c0, %c0) to (%c1, %c16) step (%c1, %c1) {
        memref.store %cst, %alloc_22[%arg12, %arg13] : memref<1x16xf32>
        scf.for %arg14 = %c0 to %c4096 step %c1 {
          %3 = memref.load %alloc_21[%arg12, %arg13, %arg14] : memref<1x16x4096xf32>
          %4 = memref.load %alloc_22[%arg12, %arg13] : memref<1x16xf32>
          %5 = arith.addf %3, %4 : f32
          memref.store %5, %alloc_22[%arg12, %arg13] : memref<1x16xf32>
        }
        scf.reduce 
      }
```

ciyongch: >  But just in case we will have an IR as below, and directly call the memref-to-cpuruntime pass. Do you think we can omit this possibility, and do not use alias analysis?

I don't think it's valid to dealloc/free the partial buffer by either `view` or `subivew` of `memref.alloc`, please check the statement here: https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/MemRef/IR/MemRefOps.td#L552-L555

crazydemo: Thanks for pointing out this reference. I will remove the alias.

ciyongch: > Why do we have such case when it is not stack-like?

Ideally, all the separated memref.alloc within the same scope should be merged into a big chunk by memref-merge pass, this fallback can be a workaround before the memref-merge patch is pulled in or in case there'll be non FILO alloc/dealloc scenarios.

Menooker: The alloc/free used to be FILO (stack like). Is there any recent optimization in upstream changing that?

> Ideally, all the separated memref.alloc within the same scope should be merged into a big chunk by memref-merge pass,

The buffer-merge pass is an optimization and can be turned off at `-O0` in the future. I suppose we should always make this pass safe to run.

lib/gc/Transforms/MemRefToCPURuntime.cpp
```diff
@@ -96,6 +97,24 @@ struct ConvertMemRefToCPURuntime
               }
             }
           }
+        } else if (isa<memref::AllocOp>(op)) {
```
Menooker, line 100: Please note that the `walk()` order may differ from the execution order. And what about the case

```
func AAA() {
     a=alloc()
     b=alloc()
     c=alloc()
     dealloc(b)
     dealloc(a)
     return c
}
```

The optimization still applies.

Menooker: Another topic is that, since we are going to handle general MLIR programs that may be hand-written by users, can your program handle this?

```
func AAA() {
     a=alloc()
     b=alloc()
     c=alloc()
     if(...) {
        dealloc(c)
     } else {
        dealloc(b)
     }
     dealloc(a)
}
```

It looks like stack-like, but it is not.

I would suggest a stricter checking:
1. first, check if an alloc is "leaked" from the scope (already done in your code in main branch)
2. for the non-leaked allocations only: check if the order of alloc/free is stack-like in the same **Region**, and ignore the leaked allocations. Usually, we can ensure the execution order in the same region.

ciyongch: > Another topic is that, since we are going to handle general MLIR programs that may be hand-written by users, can your program handle this?
> 
> ```
> func AAA() {
>      a=alloc()
>      b=alloc()
>      c=alloc()
>      if(...) {
>         dealloc(c)
>      } else {
>         dealloc(b)
>      }
>      dealloc(a)
> }
> ```
Not sure if it's a valid case or not, as it will always result in memory leak?

lib/gc/Transforms/MemRefToCPURuntime.cpp
Menooker: If we are handling dynamic shaped buffers, the buffer-merge won't work and we still need many small allocations. I would suggest in our pass pipeline that if stack-like allocator is on, we should try to make sure `BufferDeallocationPipeline` maintain a FILO pattern. If stack-like allocator is off, we can enable non-FILO pattern in BufferDeallocationPipeline

lib/gc/Transforms/MemRefToCPURuntime.cpp
Menooker: > > Another topic is that, since we are going to handle general MLIR programs that may be hand-written by users, can your program handle this?
> > ```
> > func AAA() {
> >      a=alloc()
> >      b=alloc()
> >      c=alloc()
> >      if(...) {
> >         dealloc(c)
> >      } else {
> >         dealloc(b)
> >      }
> >      dealloc(a)
> > }
> > ```
> 
> Not sure if it's a valid case or not, as it will always result in memory leak?

I think we are targeting a general MLIR compiler, so any of the user input should be handled. Users can always write code like this. At least it will not produce wrong result - if they can live with memory leak.

crazydemo:
[PR#295](https://github.com/intel/graph-compiler/pull/295) aligns the bufferization pipeline with the upstream, and can make the alloc / dealloc arranged in FILO fashion. It can fix [Issue#279](https://github.com/intel/graph-compiler/issues/279), [Issue#278](https://github.com/intel/graph-compiler/issues/278), [Issue#284](https://github.com/intel/graph-compiler/issues/284).

The current PR needs to be modified, seems `allocaOp` can be transformed to `gc runtime allocator`.

ciyongch (commented):
This change tries to resolve the FILO allocation issue, but also result in converting all the stack allocations(even with small size) into runtime allocator, maybe not that efficiency?

crazydemo:
> This change tries to resolve the FILO allocation issue, but also result in converting all the stack allocations(even with small size) into runtime allocator, maybe not that efficiency?

There're several ways to solve the alloc / dealloc style issue:
1. the current PR's solution, reuse the `PromoteBuffersToStackPass`.
2. modify the upstream `bufferization pipeline`, may need a lot discussion with the community.
3. create a GC pass, do the deallocation on GC side, may contain a large amount of logic that is redundant with upstream.
4. create a GC pass, reorganize the generated memref.alloc / memref.dealloc ops. The IR generated by the bufferization pipeline tends to have dealloc operations concentrated at the end of a region, while the order of deallocs is FIFO. Writing this pass should be kind of straightforward, but it requires extensive testing.

lib/gc/Transforms/MemRefToCPURuntime.cpp
```diff
@@ -54,48 +54,49 @@ struct AlignedAllocLowering : public OpRewritePattern<memref::AllocOp> {
...
           }
+          if (firstDeallocOp)
+            break;
+        }
+
+        // If a deallocOp was found, insert the new dealloc before it
+        if (firstDeallocOp) {
+          builder.setInsertionPoint(firstDeallocOp);
+        } else {
+          // If no deallocOp was found, insert at the end of the region before
```
Menooker, line 85: Please insert deallocop at the end of the parent alloca_scope op. It may not be the same as `op->getParentRegion()`

And I am not sure why we need to find `firstDeallocOp`?

Another issue is that, if the alloca size is very small, I think we'd better keep using the alloca instead of a runtime function. In GC v1 I remember that we have a size limit at ~128 bytes.

ciyongch: > Another issue is that, if the alloca size is very small, I think we'd better keep using the alloca instead of a runtime function. In GC v1 I remember that we have a size limit at ~128 bytes.

This is what I concern about, the current pass is trying to convert all the alloca (no matter the size) into runtime allocator which is not that efficient.

crazydemo: Thanks for the suggestion. 

1. use `getParentWithTrait<OpTrait::AutomaticAllocationScope>()` to locate the dealloc op's position.
2. need to use a while loop to find the `firstDeallocOp` to ensure the deallocation order obeys FILO.
```
// Move the insertion point back if the operation before is a deallocOp
      Operation *currentOp = builder.getInsertionPoint()->getPrevNode();
      while (currentOp && isa<cpuruntime::DeallocOp>(currentOp)) {
        // Move the insertion point before the found deallocOp
        builder.setInsertionPoint(currentOp);
        currentOp = currentOp->getPrevNode();
      }
```
3. limit size is added.

Menooker: > need to use a while loop to find the firstDeallocOp to ensure the deallocation order obeys FILO.

Please double check the IR result of `promote-allocation-to-stack`. AFAIK, there is no dealloc for alloca ops - it is "deallocated" via the end `AutomaticAllocationScope`.

crazydemo: Yes, from `promote-allocation-to-stack` to the end of `deallocation pipeline`, no dealloc will be created for alloca ops. It will be dealloced automatically at the end of `AutomaticAllocationScope`.

The above `while loop` only searches `cpuruntime::deallocOp`, to make sure the generated dealloc order as FILO.

Menooker: why do we need to check cpuruntime::deallocOp that has just been created by us in this pass?

crazydemo: When we walk through the `AutomaticAllocationScope`. If we get the below IR, the first allocated buffer `m0` will be first deallocated at the end of the scope. Then the next allocated buffer `m1` will be the next op to create the `dealloc` op, whose position should be the previous node of `dealloc %m0`
```
func.func @alloca_sequence() {
  // CHECK-LABEL: func @alloca_sequence()
  // CHECK: %[[m0:.*]] = cpuruntime.alloc() : memref<128xf32>
  // CHECK: %[[m1:.*]] = cpuruntime.alloc() : memref<128xf32>
  // CHECK: %[[m2:.*]] = cpuruntime.alloc() : memref<128xf32>
  %alloc = memref.alloca() : memref<128xf32>
  %alloc_0 = memref.alloca() : memref<128xf32>
  %alloc_1 = memref.alloca() : memref<128xf32>
  // CHECK: cpuruntime.dealloc %[[m2]] : memref<128xf32>
  // CHECK: cpuruntime.dealloc %[[m1]] : memref<128xf32>
  // CHECK: cpuruntime.dealloc %[[m0]] : memref<128xf32>
  return
}
```

crazydemo:
@ciyongch @Menooker Could you please help review this PR?

lib/gc/Transforms/MemRefToCPURuntime.cpp
```diff
@@ -54,48 +83,48 @@ struct AlignedAllocLowering : public OpRewritePattern<memref::AllocOp> {
...
-                        alias.getDefiningOp<memref::AllocOp>()) {
-                  noTransformOps.insert(allocOp);
-                }
-              }
-            }
-          }
+      funcOp.walk([&](memref::AllocaOp allocaOp) {
+        uint64_t allocSize =
+            getMemRefSizeInBytes(allocaOp.getResult().getType());
+        if (allocSize < 128) {
```
ciyongch, line 98: Please using a meaningful var (sth like STACK_ALLOC_THRESHOLD) for the number here.

crazydemo: Thanks for the advice. Fixed.

lib/gc/Transforms/MemRefToCPURuntime.cpp
crazydemo: traverse the `allocaOp` first, and then insert `deallocOp` in a reverse order now.

lib/gc/Transforms/MemRefToCPURuntime.cpp
```diff
@@ -38,9 +40,38 @@ bool hasParallelParent(Operation *op) {
   }
   return false;
 }
-struct AlignedAllocLowering : public OpRewritePattern<memref::AllocOp> {
-  using OpRewritePattern<memref::AllocOp>::OpRewritePattern;
-  LogicalResult matchAndRewrite(memref::AllocOp op,
+
+uint64_t getMemRefSizeInBytes(MemRefType memrefType) {
+  if (ShapedType::isDynamicShape(memrefType.getShape()))
```
Menooker, line 45: I guess we should use our allocator for dynamic shaped tensor by default?

crazydemo: Yes, if the memref is dynamic, then `getMemRefSizeInBytes` will return `UINT64_MAX`, shapes larger than `128Bytes` will be converted to gc runtime allocator.

Menooker: Sorry. You are right.

crazydemo:
tests are updated.

Menooker (approved):

ciyongch (approved):

PR #289, opened 2024-08-27
[Transform][Fusion] yield fused producer if necessary

Yun-Fly:
Automatically yield fused producer if it has multiple uses. E,g,

```
func.func @yield_fused_producer(%arg0: tensor<16x32x32xf32>) -> (tensor<16x32x32xf32>, tensor<16x32xf32>) {
    %cst_0 = arith.constant dense<2.000000e+00> : tensor<16x32x32xf32>
    %dest0 = tensor.empty() : tensor<16x32x32xf32>
    %0 = linalg.powf ins(%arg0, %cst_0 : tensor<16x32x32xf32>, tensor<16x32x32xf32>) outs(%dest0 : tensor<16x32x32xf32>) -> tensor<16x32x32xf32>
    %dest1 = tensor.empty() : tensor<16x32xf32>
    %1 = linalg.reduce { arith.addf } ins(%0 : tensor<16x32x32xf32>) outs(%dest1 : tensor<16x32xf32>) dimensions = [2]
    return %0, %1 : tensor<16x32x32xf32>, tensor<16x32xf32>
  }
```

When fusing `powf` into tiled consumer `reduce`, `%0` should also be yield because it has two users. Otherwise, there would exist repeat `powf` computation.

### DIFF
--- a/lib/gc/Transforms/TilingUsingInterfaceX.cpp
+++ b/lib/gc/Transforms/TilingUsingInterfaceX.cpp
@@ -21,6 +21,7 @@
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Interfaces/DestinationStyleOpInterface.h"
 #include "mlir/Interfaces/TilingInterface.h"
+#include "mlir/Transforms/RegionUtils.h"
 #include "llvm/ADT/TypeSwitch.h"
 #include "llvm/Support/Debug.h"
 #include <optional>
@@ -255,6 +256,18 @@ SmallVector<LoopLikeOpInterface> mlir::scfX::getOuterNestLoopsWhile(
   return {nestLoops.rbegin(), nestLoops.rend()};
 }
 
+/// A listener that watches which ops were erased.
+struct ErasedOpListener : public RewriterBase::Listener {
+private:
+  /// Pointers to all erased operations and blocks.
+  DenseSet<void *> erased;
+
+public:
+  ErasedOpListener() = default;
+  void notifyOperationErased(Operation *op) override { erased.insert(op); }
+  bool isErased(Operation *op) { return erased.count(op); }
+};
+
 /// Enhanced version of `tileAndFuseProducerOfSliceImpl`, which can deal with
 /// multi-level `extractSliceOp`. E.g.
 ///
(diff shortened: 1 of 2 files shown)


Yun-Fly:
This patch is also helpful to debug intermediate result when correctness check failed.

WangJialei-A (approved):

lib/gc/Transforms/TilingUsingInterfaceX.cpp
```diff
@@ -296,6 +309,57 @@ mlir::scfX::tileAndFuseProducerOfSlice(RewriterBase &rewriter,
         tileAndFuseProducerOfSliceImpl(rewriter, sliceOp, outerLoops);
     if (!fuseProducerResult)
       return std::nullopt;
+
+    // Cache old listener.
+    OpBuilder::Listener *oldListener = rewriter.getListener();
+    // Set new listener.
+    ErasedOpListener *newListener = new ErasedOpListener();
```
Menooker, line 316: Shall we call the oldListener? If `oldListener` does something important in the callback, `rewriter.setListener(newListener);` will override it. Maybe we should hold a `Listener*` in your `ErasedOpListener ` to pass down the callback to `oldListener `

lib/gc/Transforms/TilingUsingInterfaceX.cpp
```diff
@@ -296,6 +309,57 @@ mlir::scfX::tileAndFuseProducerOfSlice(RewriterBase &rewriter,
...
+                              sizes = extractSliceOp.getMixedSizes(),
+                              strides = extractSliceOp.getMixedStrides();
+    // Explicitly execute DCE.
+    (void)mlir::simplifyRegions(rewriter, {*producerOp->getParentRegion()});
+    // If fused producer has multiple users.
+    bool yieldReplacement = !newListener->isErased(producerOp);
+    // Reset to old listener.
+    rewriter.setListener(oldListener);
+    // Delete new listener.
+    delete newListener;
```
Menooker, line 334: explicit using `new` and `delete` is discouraged. Use make_unique instead.

BTW, why don't we simply use `ErasedOpListener` as a stack-local variable?

lib/gc/Transforms/TilingUsingInterfaceX.cpp
Yun-Fly: Yes, exactly. Add old listener hook now.

lib/gc/Transforms/TilingUsingInterfaceX.cpp
Yun-Fly: Sorry for this. Changed to stack-local variable.

lib/gc/Transforms/TilingUsingInterfaceX.cpp
```diff
@@ -255,6 +256,28 @@ SmallVector<LoopLikeOpInterface> mlir::scfX::getOuterNestLoopsWhile(
...
+  OpBuilder::Listener *oldListenerHook = nullptr;
+
+public:
+  ErasedOpListener() = default;
+  ErasedOpListener(OpBuilder::Listener *oldListener)
+      : oldListenerHook(oldListener) {}
+  void notifyOperationErased(Operation *op) override {
+    // Call old listener hook.
+    if (auto *oldListener =
+            dyn_cast_if_present<RewriterBase::Listener>(oldListenerHook))
```
Menooker, line 274: why do we need this cast?

Yun-Fly: `notifyOperationErased` is just the member function of `RewriterBase::Listener`, which is the sub-class of `OpBuilder::Listener`.

Menooker: Offline discussed.

Issue #293, opened 2024-08-28
OV-gc-mlir integration: some symbols not found


niuxiaog:
When running performance test of openvino-gc-mlir, the following error will appear:
```
JIT session error: Symbols not found: [ gcThreadAlignedMalloc, gcThreadAlignedFree, gcAlignedMalloc, gcAlignedFree ]
JIT invocation failed
Program aborted due to an unhandled Error:
Failed to materialize symbols: { (main, { _mlir_entry..omp_par.1, _mlir_entry..omp_par.2, _mlir_entry..omp_par.3, entry, _mlir_entry..omp_par.4, _mlir_entry..omp_par, _mlir_entry }) }
Aborted (core dumped)
```

PR #294, opened 2024-08-28
[CPU][Runtime] Set some CPURuntime funcs visibility to default

niuxiaog:
To solve issue #293.


lib/gc/ExecutionEngine/CPURuntime/MemoryPool.cpp
```diff
@@ -239,16 +239,24 @@ static thread_local FILOMemoryPool mainMemoryPool_{mainChunkSize};
 // if the current thread is a worker thread, use this pool
 static thread_local FILOMemoryPool threadMemoryPool_{threadlocalChunkSize};
 
-extern "C" void *gcAlignedMalloc(size_t sz) noexcept {
+#if defined _WIN32 || defined __CYGWIN__
```
Menooker, line 242: Maybe we should move this macro to a shared header in `CPURuntime` folder, as many cpp files will need it?

niuxiaog: Sure.

Menooker:
Suggest to co-work with @huanghaixin008 . The brgemm library also needs to export the symbols.

huanghaixin008:
> Suggest to co-work with @huanghaixin008 . The brgemm library also needs to export the symbols.

@niuxiaog  I will push my changes to this branch

include/gc/ExecutionEngine/CPURuntime/Utils.h
```diff
@@ -0,0 +1,13 @@
+//===-- Utils.h - Utils for CPURuntime --------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#if defined _WIN32 || defined __CYGWIN__
```
Menooker, line 9: Please add the include guard for this header. :)

lib/gc/ExecutionEngine/CPURuntime/MemoryPool.cpp
AndreyPavlenko: May be move it even higher? I believe it's required not only for CPURuntime. What about renaming gc_version.h to gc.h and putting the common staff there?

niuxiaog: > May be move it even higher? I believe it's required not only for CPURuntime.

What about to do this only when there is a real requirement?

> renaming gc_version.h to gc.h and putting the common staff there

The name `gc_version.h` conveys more specific information and may be a better name. The newly added `CPURuntime/Utils.h` and `gc_version.h` contains macros with totally different purpose. Currently these is no file which needs to include them both. So putting them together may not be necessary.

AndreyPavlenko: > What about to do this only when there is a real requirement?

Currently, except of CPURuntime, we need it at least for lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp and src/dnnl/dnnl_graph_compiler.cpp

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
```diff
@@ -18,11 +18,9 @@
 #include <stdexcept>
 #include <vector>
 
-#ifdef _WIN32
-#define OCL_RUNTIME_EXPORT __declspec(dllexport)
-#else
-#define OCL_RUNTIME_EXPORT
-#endif // _WIN32
+#include "gc_utils.h"
```
ciyongch, line 21: Need to update the CMakeList.txt to add the directory of newly added "gc_utils.h" to resolve the build error.

niuxiaog: Yes. Thanks for the advice! I will come back to this PR soon.

include/gc_utils.h
AndreyPavlenko, line 1: I propose moving this file into the gc folder, in order to have all the headers under the single root, and renaming to Utils.h, to be consistent with the naming style of all other headers.

lib/gc/ExecutionEngine/OpenCLRuntime/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ target_compile_options (GcOpenclRuntime PUBLIC -fexceptions -frtti)
 target_include_directories(GcOpenclRuntime PRIVATE
     ${MLIR_INCLUDE_DIRS}
     ${OpenCL_INCLUDE_DIRS}
+    ${PROJECT_SOURCE_DIR}/include
```
AndreyPavlenko, line 19: I think, you can just add GcInterface to target_link_libraries(). It should propagate all the common includes, build options, etc.

niuxiaog: Yes, that's also a reasonable way.

ciyongch:
Please rebase the PR and address the comments, then we can merge it for OV integration.

lib/gc/ExecutionEngine/OpenCLRuntime/CMakeLists.txt
ciyongch: Do you plan to change this?

niuxiaog: > Do you plan to change this?

Which way do you think is better?

Menooker: Isn't `${PROJECT_SOURCE_DIR}/include` already in include path?

niuxiaog: The `${PROJECT_SOURCE_DIR}/include` is not in include path, which causes the including error of `gc_utils.h`. But now `gc_utils.h` is moved into `/include/gc`, I will modify this now. Thanks for the reminder!

ciyongch (approved):

kurapov-peter (approved):
LGTM

Issue #288, opened 2024-08-27
Extra memref.copy introduced by bufferization


zhczhong:
```
gc-opt --gc-cpu-pipeline test.mlir
```

### `bufferization.materialize_in_destination`

`bufferization.materialize_in_destination` will be materialized as a memref.copy instead of inplace writing on the output buffer. `memref.copy %alloc_1, %arg3 : memref<128x1024xf32> to memref<128x1024xf32>` could be converted to a inplace write

```MLIR
module @fragment_name attributes {"#dlti.sys_spec" = #dlti.target_system_spec<"CPU" : #dlti.target_device_spec<#dlti.dl_entry<"tile_size", 32 : i32>, #dlti.dl_entry<"num_threads", 1 : i32>>>} {
  func.func @entry(%arg0: memref<1024x1024xf32>, %arg1: memref<128x1024xf32>, %arg2: memref<128x1024xf32>, %arg3: memref<128x1024xf32>) {
    %0 = bufferization.to_tensor %arg0 restrict : memref<1024x1024xf32>
    %1 = bufferization.to_tensor %arg1 restrict : memref<128x1024xf32>
    %2 = bufferization.to_tensor %arg2 restrict : memref<128x1024xf32>
    %3 = tensor.empty() : tensor<128x1024xf32>
    %cst = arith.constant 0.000000e+00 : f32
    %4 = linalg.fill ins(%cst : f32) outs(%3 : tensor<128x1024xf32>) -> tensor<128x1024xf32>
    %5 = linalg.matmul ins(%1, %0 : tensor<128x1024xf32>, tensor<1024x1024xf32>) outs(%4 : tensor<128x1024xf32>) -> tensor<128x1024xf32>
    %6 = tensor.empty() : tensor<128x1024xf32>
    %7 = linalg.add ins(%5, %2 : tensor<128x1024xf32>, tensor<128x1024xf32>) outs(%6 : tensor<128x1024xf32>) -> tensor<128x1024xf32>
    %8 = tensor.empty() : tensor<128x1024xf32>
    %cst_0 = arith.constant 0.000000e+00 : f32
    %9 = linalg.fill ins(%cst_0 : f32) outs(%8 : tensor<128x1024xf32>) -> tensor<128x1024xf32>
    %10 = linalg.max ins(%7, %9 : tensor<128x1024xf32>, tensor<128x1024xf32>) outs(%8 : tensor<128x1024xf32>) -> tensor<128x1024xf32>
[...]

PR #295, opened 2024-08-28
Align the bufferization pipeline according to the upstream doc

zhczhong:
Fix part of issue https://github.com/intel/graph-compiler/issues/288

`empty_tensor_elimination` could eliminate the copy introduced by `bufferization.materialize_in_destination`. And adjust the pipeline according to the doc(https://mlir.llvm.org/docs/Bufferization/#overview).

### DIFF
--- a/lib/gc/Transforms/Pipeline.cpp
+++ b/lib/gc/Transforms/Pipeline.cpp
@@ -92,6 +92,7 @@ void populateVectorPasses(mlir::OpPassManager &pm) {
 
 // scf + arith + math + vector + memref + linalg.brgemm
 void populateBufferizationPasses(mlir::OpPassManager &pm) {
+  pm.addPass(bufferization::createEmptyTensorEliminationPass());
   bufferization::OneShotBufferizationOptions options;
   options.bufferizeFunctionBoundaries = true;
   options.setFunctionBoundaryTypeConversion(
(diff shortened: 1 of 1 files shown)


lib/gc/Transforms/Pipeline.cpp
```diff
@@ -92,21 +92,27 @@ void populateVectorPasses(mlir::OpPassManager &pm) {
...
 void populateBufferizationPasses(mlir::OpPassManager &pm) {
+  pm.addPass(bufferization::createEmptyTensorEliminationPass());
   bufferization::OneShotBufferizationOptions options;
   options.bufferizeFunctionBoundaries = true;
   options.setFunctionBoundaryTypeConversion(
       bufferization::LayoutMapOption::IdentityLayoutMap);
   pm.addPass(bufferization::createOneShotBufferizePass(options));
   pm.addPass(createCanonicalizerPass());
   pm.addPass(createCSEPass());
+  pm.addPass(createCanonicalizerPass());
```
ZhennanQin, line 103: L101 is already CanonicalizerPass, do we really need to do it again after CSE?

zhczhong: Yes, the first `CanonicalizerPass` is mainly used to canonicalize the `scf.for` and `scf.forall` to the canonical form in memref, exposing more `CSE` optimization opportunities. The `CSE` pass is used to eliminate the common `memref.subview` so there will be some useless memref.copy `memref.copy %a, %a`. The last `CanonicalizerPass` is used to eliminate the useless `memref.copy` which copies from A to A like `memref.copy %a, %a`

lib/gc/Transforms/Pipeline.cpp
```diff
@@ -92,21 +92,27 @@ void populateVectorPasses(mlir::OpPassManager &pm) {
...
   pm.addPass(bufferization::createBufferResultsToOutParamsPass(opt));
-  // todo: buffer schedule pass
-  // todo: Need to improve this pass to support nested parallel.
-  pm.addNestedPass<func::FuncOp>(bufferization::createBufferHoistingPass());
-  pm.addNestedPass<func::FuncOp>(bufferization::createBufferLoopHoistingPass());
-  pm.addNestedPass<func::FuncOp>(bufferization::createBufferDeallocationPass());
+  pm.addPass(bufferization::createDropEquivalentBufferResultsPass());
+  pm.addNestedPass<func::FuncOp>(
+      bufferization::createPromoteBuffersToStackPass());
+  mlir::bufferization::BufferDeallocationPipelineOptions deallocOption;
```
ciyongch, line 114: nit: no need to add `mlir` namespace here.
BTW, could you also put a note/link to the upstream bufferization pipeline?

Menooker (approved):

lib/gc/Transforms/Pipeline.cpp
zhczhong: Thanks for the advice! the mlir namespace has been removed and the note has been added

ciyongch (approved):

ZhennanQin (approved):

PR #309, opened 2024-09-02
Refactor the deep tile matmul config and skip the single-iteration loop generation

zhczhong:
Track: https://github.com/intel/graph-compiler/issues/288

1. The extra memref.copy is caused by write-after-write conflict because the canonicalization pass eliminates the single-iteration loop but preserves the extracted slice and insert slice. So here skip generating them when they are single-iteration loop.
2. Refactor the config. Infer dimType according to the contractionOpInterface.
3. introduce padding cost, which minimizes the cost on padding and use divisible block if possible


lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -278,6 +281,13 @@ prepareConfigCandidates(Operation *root, CPUTargetDescriptionAnalysis &sysDesc,
           : getCandidate((uint32_t)shape[2],
                          shape[2] >= noSmallBlockNeedThreshold ? 8U : 1U, 256U);
 
+  if (allowIndivisibleInnerblock) {
+    innerMostKBlockCandidates = {16, 32, 64};
+    innerMostNBlockCandidates = {16, 32, 64};
+    NBlockCandidates = innerMostNBlockCandidates;
+    KBlockCandidates = innerMostKBlockCandidates;
```
yifeizh2, line 288: So after the change here, innermost Kblock will only be one of 16/32/64 if `allowIndivisibleInnerblock` is true?

zhczhong: Yes, your understanding is correct

lib/gc/Transforms/DeepTileContractionOp.cpp
```diff
@@ -107,6 +107,16 @@ tensorViewRankedTensor(RewriterBase &rewriter, RankedTensorType outTensorType,
   return result;
 }
 
+// Check if the loop is dummy loop(has only one iteration)
+bool isDummyLoop(LoopLikeOpInterface loop) {
```
ciyongch, line 111: Seems this is another impl besides "promoteIfSingleIteration()"?

lib/gc/Transforms/DeepTileContractionOp.cpp
```diff
@@ -260,31 +270,34 @@ static void getMatmulParallelDims(linalg::LinalgOp linalgOp,
...
-      Value viewResult = tensorViewRankedTensor(
-          rewriter, targetTensor, newExtractSliceOp->getResult(0));
-      rewriter.replaceOp(extractSlice, viewResult);
-    } else {
-      rewriter.replaceOp(extractSlice, newExtractSliceOp);
-    }
+    rewriter.replaceOp(op, currentOp);
+  }
+
+  if (shrinDimNum > 0) {
```
ciyongch, line 291: nit: shrinDimNum->shrinkDimNum?

lib/gc/Transforms/DeepTileContractionOp.cpp
zhczhong: Yes, the purpose is similar but the `promoteIfSingleIteration()` erase the single iteration loop only after the loop is created. The extract/insert slices will be preserved, which causes the WAW conflict in bufferization and introduces more `memref.copy`. Here skip generating the single iteration loop and its corresponding extract/insert_slice op in the creation  phase once detecting the single iteration.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -55,9 +55,10 @@ bool validateConfig(const MatmulConfig &cfg) {
 std::vector<uint32_t>
 getCandidate(uint32_t num, uint32_t floor,
              uint32_t ceil = std::numeric_limits<uint32_t>::max()) {
+  int defaultBlock = 32;
   // factor
   std::vector<uint32_t> candidates;
-  uint32_t upperbound = std::min(num, ceil);
+  uint32_t upperbound = std::min(llvm::divideCeil(num, defaultBlock), ceil);
```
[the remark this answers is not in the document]
zhczhong: Thanks for pointing it out! should be `llvm::divideCeil(num, defaultBlock) * defaultBlock`

lib/gc/Transforms/DeepTileContractionOp.cpp
zhczhong: Fixed

include/gc/Analysis/MatmulConfigAnalysis.h
```diff
@@ -47,15 +47,67 @@ inline SmallVector<unsigned> extractDimTypeIdx(ArrayRef<DimType> tyList,
   return idxList;
 }
 
+inline void getDimTypeFromIterators(linalg::LinalgOp linalgOp,
+                                    SmallVectorImpl<DimType> &dimTypes) {
+  SmallVector<utils::IteratorType> iteratorTypes =
```
yifeizh2, line 52: Can you explicitly specify `mlir::utils::IteratorType` here?

lib/gc/Transforms/DeepTileContractionOp.cpp
ciyongch: I see, thanks for the explanation.

include/gc/Analysis/MatmulConfigAnalysis.h
zhczhong: add `mlir` namespace here

ciyongch:
Overall LGTM, do you think we should add a single simple case to cover the case of skipping single-iteration loop generation?

zhczhong:
> Overall LGTM, do you think we should add a single simple case to cover the case of skipping single-iteration loop generation?

This test could cover this case. The file check will check two `scf.forall` before this PR but the single-iteration `scf.forall` will be skipped now.

https://github.com/intel/graph-compiler/blob/613a333ee5e14731feeca40b54c57adb07ca7edc/test/mlir/test/gc/Transforms/deepTileContractionNamedOp.mlir#L127

ciyongch:
Please help to rebase the code base, then we can merge it.

zhczhong:
> Please help to rebase the code base, then we can merge it.

The code has been rebased now

ciyongch (approved):

yifeizh2 (approved):

Issue #315, opened 2024-09-03
`IterativeTilingAndFusion` failed with latest config enhanced deep tile matmul


yifeizh2:
Failed case (IR after deep tile matmul):
```
#map = affine_map<(d0) -> (d0 * 32)>
module {
  func.func @main_entry(%arg0: tensor<128x512xbf16>, %arg1: tensor<512x1024xbf16>, %arg2: tensor<1024xbf16>) -> tensor<128x1024xbf16> attributes {llvm.emit_c_interface} {
    %c32 = arith.constant 32 : index
    %c64 = arith.constant 64 : index
    %c1 = arith.constant 1 : index
    %c2 = arith.constant 2 : index
    %c0 = arith.constant 0 : index
    %cst = arith.constant dense<0.000000e+00> : tensor<128x1024xbf16>
    %0 = tensor.empty() : tensor<128x1024xbf16>
    %1 = tensor.empty() : tensor<32x16x32x32xbf16>
    %pack = tensor.pack %arg1 outer_dims_perm = [1, 0] inner_dims_pos = [0, 1] inner_tiles = [32, 32] into %1 : tensor<512x1024xbf16> -> tensor<32x16x32x32xbf16>
    %2 = tensor.empty() : tensor<32x16x16x32x2xbf16>
    %pack_0 = tensor.pack %pack inner_dims_pos = [2] inner_tiles = [2] into %2 : tensor<32x16x32x32xbf16> -> tensor<32x16x16x32x2xbf16>
    %3 = scf.forall (%arg3) = (0) to (128) step (64) shared_outs(%arg4 = %0) -> (tensor<128x1024xbf16>) {
      %extracted_slice = tensor.extract_slice %arg0[%arg3, 0] [64, 512] [1, 1] : tensor<128x512xbf16> to tensor<64x512xbf16>
      %extracted_slice_1 = tensor.extract_slice %arg4[%arg3, 0] [64, 1024] [1, 1] : tensor<128x1024xbf16> to tensor<64x1024xbf16>
      %9 = scf.forall (%arg5) = (0) to (32) step (2) shared_outs(%arg6 = %extracted_slice_1) -> (tensor<64x1024xbf16>) {
        %10 = affine.apply #map(%arg5)
[...]

PR #316, opened 2024-09-03
[Transform][Fusion] fix `whileProducerOutOfLoopBlock` logic

Yun-Fly:
Track #315.

BTW, the root cause is that upstream function `addInitOperandsToLoopNest` assumes nested loops must be `scf.for` instead of `scf.forall`. It may take more time to fix it upstream.


yifeizh2:
Tested on the original failed case. The issue is resolved.

yifeizh2 (approved):

zhczhong (approved):

PR #334, opened 2024-09-10
Reuse code, fix Coverity scans

kwasd:
This fixes Coverity scans and unifies CI code.


kwasd:
> I recommend using script instead of actions, in this way developers can use the same env locally

Will implement in a separate PR

Issue #323, opened 2024-09-05
Performance regression caused by read lock in brgemm


zhczhong:
According to the perf test from @yifeizh2, the read lock will bring 10-20% performance regression https://github.com/intel/graph-compiler/blob/bc0014b12bb58b554277ab74c248535fff8c0cec/lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp#L145

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/zhicongz/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/zhicongz/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">
<span data-teams=true>

  data type| bs  | hidden_list  | without lock | with lock |  
-- | -- | -- | -- | -- | --
bf16 | 128 | 16x512 | 0.021597 | 0.01967 | 109.79%
bf16 | 128 | 512x256 | 0.025172 | 0.029239 | 86.09%
bf16 | 128 | 256x128 | 0.015969 | 0.026034 | 61.34%
bf16 | 128 | 512x1024 | 0.046416 | 0.055459 | 83.69%
bf16 | 128 | 1024x1024 | 0.082655 | 0.089712 | 92.13%
bf16 | 128 | 1024x512 | 0.066204 | 0.066523 | 99.52%
bf16 | 128 | 512x256 | 0.025992 | 0.029516 | 88.06%

</body>

</html>

Menooker:
` read_lock_guard_t g(g_brgemm_lock); `

The reader-writer lock will result in serialization in memory even all users of the lock are readers. I think we might need to re-design this, to make it totally lock-free.

crazydemo:
> `read_lock_guard_t g(g_brgemm_lock);`
> 
> The reader-writer lock will result in serialization in memory even all users of the lock are readers. I think we might need to re-design this, to make it totally lock-free.

[...]

PR #350, opened 2024-09-20
add thread local cache for brgemm

crazydemo:
Tracking [Issue#323](https://github.com/intel/graph-compiler/issues/323)

### DIFF
--- a/lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
+++ b/lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
@@ -53,56 +53,14 @@ using read_lock_guard_t = std::shared_lock<std::shared_mutex>;
 using write_lock_guard_t = std::unique_lock<std::shared_mutex>;
 static std::shared_mutex g_brgemm_lock;
 
-static std::vector<brgemm_desc_t> g_brgemm_desc_list;
-static std::vector<brgemm_kernel_t *> g_brgemm_kernel_list;
-static std::vector<std::unique_ptr<char[]>> g_brgemm_palette;
-
 struct brgemm_cache_info_t {
-  std::shared_ptr<brgemm_desc_t> desc;
-  std::shared_ptr<brgemm_kernel_t> kernel;
-  std::shared_ptr<char> palette;
-
-  brgemm_cache_info_t() = default;
-  brgemm_cache_info_t(brgemm_desc_t *d, brgemm_kernel_t *k, char *p)
-      : desc(d), kernel(k), palette(p) {}
-  brgemm_cache_info_t &operator=(const brgemm_cache_info_t &other) {
-    if (this != &other) {
-      desc = other.desc;
-      kernel = other.kernel;
-      palette = other.palette;
-    }
-    return *this;
-  }
+  brgemm_desc_t desc;
+  brgemm_kernel_t *kernel;
+  std::shared_ptr<char[]> palette;
 };
 
-class brgemm_cache_manager {
-public:
-  static brgemm_cache_manager &getInstance() {
-    static thread_local brgemm_cache_manager instance;
-    return instance;
-  }
-
-  void insertOrUpdate(int64_t key, const brgemm_cache_info_t &info) {
-    cache_[key] = info;
-  }
-
-  bool tryGet(int64_t key, brgemm_cache_info_t &info) {
-    auto it = cache_.find(key);
-    if (it != cache_.end()) {
-      info = it->second;
-      return true;
-    }
-    return false;
-  }
-
-private:
-  brgemm_cache_manager() {}
-
-  brgemm_cache_manager(const brgemm_cache_manager &) = delete;
-  brgemm_cache_manager &operator=(const brgemm_cache_manager &) = delete;
-
-  std::unordered_map<int64_t, brgemm_cache_info_t> cache_;
-};
+static std::vector<brgemm_cache_info_t> g_cache;
+static thread_local std::unordered_map<int64_t, brgemm_cache_info_t> tl_cache;
 
 // TODO(haixin): use syscall to determine page size?
 static constexpr size_t SCRATCH_SIZE = 2 * 4096;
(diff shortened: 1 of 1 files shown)


lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
```diff
@@ -137,24 +150,38 @@ void dnnl_brgemm_tilerelease() {
 void dnnl_brgemm_execute(int64_t kernel_idx, void *A, uint64_t A_offset,
                          void *B, uint64_t B_offset, void *C, uint64_t C_offset,
                          int num) {
+  auto it = tl_cache.find(kernel_idx);
```
ZhennanQin, line 153: It's better not define tl_cache as global static, define it here as a function static.

lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
```diff
@@ -137,24 +150,38 @@ void dnnl_brgemm_tilerelease() {
...
-  size_t B_offset_in_bytes;
-  size_t C_offset_in_bytes;
-  {
+  brgemm_desc_t *desc_ptr = nullptr;
+
+  if (it != tl_cache.end()) {
+    desc_ptr = &it->second.desc;
+    kernel = it->second.kernel;
+  } else {
     read_lock_guard_t g(g_brgemm_lock);
```
ZhennanQin, line 161: Since it's thread local, do we still need this lock?

lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
```diff
@@ -57,6 +57,14 @@ static std::vector<brgemm_desc_t> g_brgemm_desc_list;
 static std::vector<brgemm_kernel_t *> g_brgemm_kernel_list;
 static std::vector<std::unique_ptr<char[]>> g_brgemm_palette;
 
+struct brgemmInfo {
+  brgemm_desc_t desc;
+  brgemm_kernel_t *kernel;
+  char *palette;
```
huanghaixin008, line 63: I suggest not to store raw pointer managed by smart pointer, maybe change to `shared_ptr`?

lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
```diff
@@ -57,6 +57,14 @@ static std::vector<brgemm_desc_t> g_brgemm_desc_list;
 static std::vector<brgemm_kernel_t *> g_brgemm_kernel_list;
 static std::vector<std::unique_ptr<char[]>> g_brgemm_palette;
 
+struct brgemmInfo {
```
huanghaixin008, line 60: Suggest change the name to `brgemm_cache_info_t` to align with naming convention in this file, and avoid confusion

lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
```diff
@@ -57,6 +57,53 @@ static std::vector<brgemm_desc_t> g_brgemm_desc_list;
 static std::vector<brgemm_kernel_t *> g_brgemm_kernel_list;
 static std::vector<std::unique_ptr<char[]>> g_brgemm_palette;
 
+struct brgemm_cache_info_t {
+  std::shared_ptr<brgemm_desc_t> desc;
+  std::shared_ptr<brgemm_kernel_t> kernel;
+  std::shared_ptr<char> palette;
```
huanghaixin008, line 63: we could use `shared_ptr<char[]>` for `palette`, and for `desc` and `kernel` we don't need to change, since they are not managed by smart ptr, and storing ptr to vector elements is dangerous as well

lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
```diff
@@ -57,6 +57,53 @@ static std::vector<brgemm_desc_t> g_brgemm_desc_list;
...
+
+  brgemm_cache_info_t() = default;
+  brgemm_cache_info_t(brgemm_desc_t *d, brgemm_kernel_t *k, char *p)
```
huanghaixin008, line 66: ideally we need to change the `unique_ptr` in global palette pool to `shared_ptr` as well, and pass the `shared_ptr` of palette here for construction

lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
yifeizh2, line 63: As we created `brgemm_cache_info_t` to store desc/kernel/palette together for thread locally, would it be a better manner to also use `brgemm_cache_info_t` for global management?

huanghaixin008: I think it's a good idea, we can unify the struct used in both thead-local and global

lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
```diff
@@ -105,25 +152,30 @@ int64_t dnnl_brgemm_dispatch(int64_t M, int64_t N, int64_t K, int64_t LDA,
...
     read_lock_guard_t g(g_brgemm_lock);
-    assert(kernel_idx >= 0 && kernel_idx < (int64_t)g_brgemm_desc_list.size() &&
+    assert(kernel_idx < (int64_t)g_brgemm_desc_list.size() &&
            "Invalid kernel handler");
-    brgemm_desc_t &desc = g_brgemm_desc_list[kernel_idx];
-    if (!desc.is_tmm) {
-      return;
-    }
-    palette_buffer = g_brgemm_palette[kernel_idx].get();
+    info = {&g_brgemm_desc_list[kernel_idx], g_brgemm_kernel_list[kernel_idx],
```
zhczhong, line 166: It is not safe to assign the raw pointer to the shared_ptr in struct which will release the pointer when the `ref_count == 0`

crazydemo: fixed.

lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
crazydemo: fixed.

lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
crazydemo: fixed.

yifeizh2 (approved):

lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
crazydemo: thanks for the advice, fixed.

lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
crazydemo: when the target brgemm kernel is not found in thread_local cache, we still need to lock the global cache to get the target brgemm.

lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
```diff
@@ -93,33 +102,33 @@ int64_t dnnl_brgemm_dispatch(int64_t M, int64_t N, int64_t K, int64_t LDA,
   brgemm_desc_set_attr(&desc, dnnl_attrs);
 
   // TODO(haixin): Reuse identical palettes across kernels
-  char *palette_buffer = nullptr;
+  std::shared_ptr<char[]> palette_buffer(new char[PALETTE_SIZE],
+                                         std::default_delete<char[]>());
```
huanghaixin008, line 106: We only need to `new` palette buffer when `desc.is_tmm` is `true`.

crazydemo: fixed.

huanghaixin008 (approved):

zhczhong (approved):

lib/gc/ExecutionEngine/CPURuntime/Microkernel/BrgemmOnednn.cpp
```diff
@@ -53,15 +53,24 @@ using read_lock_guard_t = std::shared_lock<std::shared_mutex>;
...
+
+static std::vector<brgemm_cache_info_t> g_cache;
 
 // TODO(haixin): use syscall to determine page size?
 static constexpr size_t SCRATCH_SIZE = 2 * 4096;
 // TODO(haixin): need to use custom thread management for scratch in the future?
 static thread_local char scratch[SCRATCH_SIZE] = {0};
 
+static std::unordered_map<int64_t, brgemm_cache_info_t> &get_tl_cache() {
+  thread_local std::unordered_map<int64_t, brgemm_cache_info_t> tl_cache;
```
Menooker, line 70: Sorry I am late for the party. Can we use std::vector for better performance?

ciyongch: The "key" here might not be contiguous?

Menooker: Haixin originally used a vector to hold the kernels. I think he tried to make them contiguous. Need to double check that.

huanghaixin008: In global, it's contiguous, but in thread local cache it might be not.
But I think we can still use `vector` for thread local cache, with empty 'hole's inside the vector.
Using unordered_map indeed would bring some extra cost.

crazydemo: The previous design was able to use a vector for access because there was only a single global cache storing the BRGEMM information. This PR introduces a new thread-local cache, and the indices in this cache may not necessarily align with those in the global cache.

Menooker: I think it still profitable to use a vector. It is contiguous in memory and in most of time, it should be dense (will it be common when a thread calls brgemm A, and another calls brgemm B?) Please note that std::unordered_map is slow and space-consuming. It stores k-v for each pair and the pairs are stored in a linked list. That is at least 3 times of the space of a vector.

PR #100, opened 2024-05-27
[Transform][Vectorize] lower linalg named op to math/arith and canonicalize vector with physical vector

BRUCE11111:
Tracking [issue 55](https://github.com/intel/graph-compiler/issues/55)

Todo:
 1. - [x] lower linalg named operations and some tensor operations to math/arith.
 2. - [x] Simple operations like elementwise operations lower to physical vector.
 3. - [x] Simple operations like elementwise operations fusion.
 4. - [ ] Optimize special operations like reduce.
 5. - [ ] Special operations fusion.

PR #363, opened 2024-09-27
Update llvm commit

BRUCE11111:
Need to fix ci segmentation fault error in [PR100](https://github.com/intel/graph-compiler/pull/100).

When we need to determine whether `linalg.generic` is a matmul operation, we will use upstream `isContractionBody` function, but it will throw seg fault error due to nullptr pointer. The null pointer issue has been fixed in this llvm commit.


ciyongch:
Better to describe the reason why we need to update the commit it, like this upstream contains a specific fix/update on certain feature?

BRUCE11111:
> Better to describe the reason why we need to update the commit it, like this upstream contains a specific fix/update on certain feature?

Okay~ Added detailed explanation in the PR.

ciyongch (approved):

zhczhong (approved):

Issue #375, opened 2024-10-09
Invalid config encountered at tuning stage for deepTiledMatmul


yifeizh2:
The following matmul config will lead to segfault at workload execution stage:

```
module attributes {dlti.target_system_spec = #dlti.target_system_spec<"CPU" : #dlti.target_device_spec<#dlti.dl_entry<"L1_cache_size_in_bytes", 49152 : ui32>, #dlti.dl_entry<"L2_cache_size_in_bytes", 2097152 : ui64>, #dlti.dl_entry<"L3_cache_size_in_bytes", 110100480 : ui64>, #dlti.dl_entry<"num_threads", 56 : i32>, #dlti.dl_entry<"max_vector_width", 512 : i64>>>} {
  func.func @entry(%arg0: tensor<128x4096xbf16>, %arg1: tensor<344x128x16x32x2xbf16>, %arg2: tensor<11008xbf16>) -> tensor<128x11008xbf16> attributes {llvm.emit_c_interface} {
    %cst = arith.constant dense<0.000000e+00> : tensor<128x11008xbf16>
    %cst_0 = arith.constant 0.000000e+00 : bf16
    %0 = tensor.empty() : tensor<128x11008xbf16>
    %1 = linalg.fill ins(%cst_0 : bf16) outs(%0 : tensor<128x11008xbf16>) -> tensor<128x11008xbf16>
[...]

PR #376, opened 2024-10-09
Add config validity check

yifeizh2:
Add matmul config validity check to avoid invalid configs.

Resolves issue #375 and #377.

### DIFF
--- a/lib/gc/Analysis/MatmulConfigAnalysis.cpp
+++ b/lib/gc/Analysis/MatmulConfigAnalysis.cpp
@@ -37,7 +37,7 @@ static llvm::raw_ostream &operator<<(llvm::raw_ostream &ss,
   return ss;
 }
 
-bool validateConfig(const MatmulConfig &cfg) {
+bool validateConfig(const MatmulConfig &cfg, ArrayRef<uint32_t> shape) {
   if (cfg.MThreads <= 0 || cfg.NThreads <= 0 || cfg.KThreads <= 0 ||
       cfg.MBlock <= 0 || cfg.NBlock <= 0 || cfg.KBlock <= 0 ||
       cfg.innerMostMBlock <= 0 || cfg.innerMostNBlock <= 0 ||
@@ -47,6 +47,12 @@ bool validateConfig(const MatmulConfig &cfg) {
       cfg.NBlock % cfg.innerMostNBlock != 0 ||
       cfg.KBlock % cfg.innerMostKBlock != 0)
     return false;
+  if (!shape.empty()) {
+    // KThreads will not shrink automatically
+    // K is shape[2]
+    if (llvm::divideCeil(shape[2], cfg.KBlock) < cfg.KThreads)
+      return false;
+  }
   return true;
 }
 
@@ -179,7 +185,7 @@ double dynamicBufferizationCost(linalg::LinalgOp &linalgOp,
                                 ArrayRef<uint32_t> shape,
                                 const MatmulConfig &config,
                                 CPUTargetDescriptionAnalysis &sysDesc) {
-  assert(validateConfig(config) && "config is invalid");
+  assert(validateConfig(config, shape) && "config is invalid");
   assert(shape.size() >= 3 && "shape.size() should >= 3");
   uint32_t M = shape[0], N = shape[1];
   double cost = 0;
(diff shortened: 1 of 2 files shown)


yifeizh2:
I will further add unit test or local check with benchgc. Currently, we also require the tuner side to reject such invalid config @xurui1995

xurui1995:
> I will further add unit test or local check with benchgc. Currently, we also require the tuner side to reject such invalid config @xurui1995

I will add this validating rule on python side

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -393,10 +400,10 @@ bool readConfigFromAttrs(MatmulConfig &config, ArrayRef<NamedAttribute> attrs) {
       cfgItemCnt++;
     }
   }
-  if (validateConfig(config)) {
+  if (validateConfig(config, shape)) {
     return cfgItemCnt == 9;
   } else {
-    LLVM_DEBUG(llvm::dbgs() << "The predefined config is invalid\n");
+    assert(validateConfig(config, shape) && "config is invalid");
```
yifeizh2, line 406: @zhczhong Replaced the original debug info with assertion to avoid invalid config being replaced with a valid one, which potentially will lead to error in tuning result.

WangJialei-A:
> > I will further add unit test or local check with benchgc. Currently, we also require the tuner side to reject such invalid config @xurui1995
> 
> I will add this validating rule on python side

It is not reasonable to add special check on benchgc side and we do not need to maintain two implementations. Throw an exception or error code to told the caller that the config is invalid is a better way, I think.

yifeizh2:
> > > I will further add unit test or local check with benchgc. Currently, we also require the tuner side to reject such invalid config @xurui1995
> > 
> > 
> > I will add this validating rule on python side
> 
> It is not reasonable to add special check on benchgc side and we do not need to maintain two implementations. Throw an exception or error code to told the caller that the config is invalid is a better way, I think.

Just replaced the original debug message with an assertion.

WangJialei-A:
> > > > I will further add unit test or local check with benchgc. Currently, we also require the tuner side to reject such invalid config @xurui1995
> > > 
> > > 
> > > I will add this validating rule on python side
> > 
> > 
> > It is not reasonable to add special check on benchgc side and we do not need to maintain two implementations. Throw an exception or error code to told the caller that the config is invalid is a better way, I think.
> 
> Will add that too. And I still prefer adding a validation rule on python side which can help reduce tuning time

From my perspective, it will be better if function `validateConfig` is exposed in python binding.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -393,10 +400,15 @@ bool readConfigFromAttrs(MatmulConfig &config, ArrayRef<NamedAttribute> attrs) {
...
-    LLVM_DEBUG(llvm::dbgs() << "The predefined config is invalid\n");
+  if (cfgItemCnt != 9) {
+    LLVM_DEBUG(llvm::dbgs() << "The predefined matmul config is incomplete. "
+                               "Default matmul config will be set.\n");
+    return false;
+  }
+  if (validateConfig(config, shape))
+    return true;
+  else {
+    assert(0 && "config is invalid");
```
ciyongch, line 411: Better to use llvm::dbgs() to give the warning when there's invalid config, `assert` only throw error in debug mode.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
```diff
@@ -47,6 +47,12 @@ bool validateConfig(const MatmulConfig &cfg) {
       cfg.NBlock % cfg.innerMostNBlock != 0 ||
       cfg.KBlock % cfg.innerMostKBlock != 0)
     return false;
+  if (!shape.empty()) {
```
ciyongch, line 50: If this is the case that generated by heuristic rule, we also need to update the logic there?
BTW, seems it only use the `K` dim size, no need to pass the entire `shape` here?

lib/gc/Analysis/MatmulConfigAnalysis.cpp
yifeizh2: > Better to use llvm::dbgs() to give the warning when there's invalid config, `assert` only throw error in debug mode.

I will update to giving both `assert` and `llvm::dbgs()`.

lib/gc/Analysis/MatmulConfigAnalysis.cpp
yifeizh2: > If this is the case that generated by heuristic rule, we also need to update the logic there? BTW, seems it only use the `K` dim size, no need to pass the entire `shape` here?

Heuristic rule will not generate this kind of invalid config. This issue is exposed by tuner side generated config. 
The reason for only checking `K` dim size is because, only the `KThreads` value will not shrink automatically while `M` and `N` would (which is an upstream limitation). I currently passed the entire `shape` here in case we further need to check shape related constraint for `M` and `N` in the future.

ciyongch: I see. I would suggest we only pass the things that is necessary as you mentioned on `M` and `N`, anyway you can keep it as is.

ciyongch (approved):

PR #384, opened 2024-10-21
Fix compile.sh for a clean build

kurapov-peter:
A clean build currently fails with:
```
error: pathspec '.' did not match any file(s) known to git
```

This patch fixes the problem which is creating llvm source folder unconditionally before checking for its existence.


dchigarev (approved):

AndreyPavlenko (approved):

leshikus (commented):
LGTM

PR #389, opened 2024-10-25
Add XeVM block store/load ops definition

kurapov-peter:
I've split the single cache control record into multiple of the same class to add some flexibility for changes to cache control policies and their number. It makes the ops fields a bit more explicit but verification (and other) logic will now be isolated for each cache level.

Depends on #387.


dchigarev (approved):
is this dialect going to replace xegpu? Or it will be somewhere higher/lower in the pipeline?

kurapov-peter:
This is the one below xegpu, at the same level as llvm.

AndreyPavlenko (approved):

PR #390, opened 2024-10-25
[GPU] Add 2d load/store ops validation

kurapov-peter:
This adds the necessary validation for HW parameters of the memory operations and diagnostic tests.

Depends on #389

PR #395, opened 2024-10-28
[GPU] XeVM to LLVM pass structure

kurapov-peter:
This adds the lowering pass boilerplate for `convert-xevm-to-llvm`.

Depends on #390.

### DIFF
--- a/cmake/functions.cmake
+++ b/cmake/functions.cmake
@@ -119,6 +119,16 @@ function(gc_add_mlir_dialect_library name)
     endif()
 endfunction()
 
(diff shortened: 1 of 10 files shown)


dchigarev (approved):
We might also want to move some of the passes from 'Transforms' to 'Conversion' at some point in the future, right? (linalg-to-xegpu, gpu-to-gpuocl, lower-to-tile-vector, convert-memref-to-cpuruntime, ...)

AndreyPavlenko (approved):

PR #413, opened 2024-11-26
[GpuOclRuntime] Avoid error when iterating over non-gpu devices

akroviakov:
Consider these devices visible to OpenCL (CPU and GPU):
```
Platform: Intel(R) OpenCL
  Device: Intel(R) Xeon(R) Gold 6438Y+
Platform: Intel(R) OpenCL Graphics
  Device: Intel(R) Data Center GPU Max 1100
```
If you try to run the `mlp.mlir` example in the current `main`, it would display an error (`-1`) for CPU, but would still run the test. The error is displayed due to reading __all__ devices using `CL_DEVICE_TYPE_GPU`. Instead, we can first read all devices using `CL_DEVICE_TYPE_ALL` and then use `CL_DEVICE_TYPE` to select GPUs, this way, no error is displayed.

### DIFF
--- a/lib/gc/ExecutionEngine/GPURuntime/ocl/GpuOclRuntime.cpp
+++ b/lib/gc/ExecutionEngine/GPURuntime/ocl/GpuOclRuntime.cpp
@@ -352,7 +352,7 @@ OclRuntime::gcIntelDevices(size_t max) {
 
   for (auto platform : platforms) {
     cl_uint numDevices;
-    err = clGetDeviceIDs(platform, CL_DEVICE_TYPE_GPU, 0, nullptr, &numDevices);
+    err = clGetDeviceIDs(platform, CL_DEVICE_TYPE_ALL, 0, nullptr, &numDevices);
     if (err != CL_SUCCESS) {
       gcLogE("Failed to get the number of devices on the platform.", platform,
              " Error: ", err);
@@ -363,7 +363,7 @@ OclRuntime::gcIntelDevices(size_t max) {
     }
 
     SmallVector<cl_device_id> devices(numDevices);
-    err = clGetDeviceIDs(platform, CL_DEVICE_TYPE_GPU, numDevices,
+    err = clGetDeviceIDs(platform, CL_DEVICE_TYPE_ALL, numDevices,
                          devices.data(), nullptr);
     if (err != CL_SUCCESS) {
       gcLogE("Failed to get the device ids on the platform ", platform,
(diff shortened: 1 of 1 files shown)


AndreyPavlenko:
> Consider these devices visible to OpenCL (CPU and GPU):
> 
> ```
> Platform: Intel(R) OpenCL
>   Device: Intel(R) Xeon(R) Gold 6438Y+
> Platform: Intel(R) OpenCL Graphics
>   Device: Intel(R) Data Center GPU Max 1100
> ```
> 
> If you try to run the `mlp.mlir` example in the current `main`, it would display an error (`-1`) for CPU, but would still run the test. The error is displayed due to reading **all** devices using `CL_DEVICE_TYPE_GPU`. Instead, we can first read all devices using `CL_DEVICE_TYPE_ALL` and then use `CL_DEVICE_TYPE` to select GPUs, this way, no error is displayed.

AndreyPavlenko:
Sorry, I closed it accidently.

kurapov-peter:
What's the use case for running CPU through opencl anyway? The runtime is not supposed to be called during CPU execution.

akroviakov:
I build like this:
```
git clone ...
./scripts/compile.sh --dev --imex
cd build
cmake --build . --target gc-check
```

then I run `mlp.mlir` like this:
```
/home/.../graph-compiler/build/bin/gc-gpu-runner --shared-libs=/home/.../graph-compiler/externals/llvm-project/build/lib/libmlir_runner_utils.so /home/.../graph-compiler/test/mlir/test/gc/gpu-runner/mlp.mlir
```

Assuming I am using gpu-runner in a way I am not supposed to, what would be the correct command to run a single arbitrary `.mlir` test for GPU then?

AndreyPavlenko:
> Assuming I am using gpu-runner in a way I am not supposed to, what would be the correct command to run a single arbitrary `.mlir` test for GPU then?

You are using it in the right way and the same command works for me:
```
$ ./bin/gc-gpu-runner --shared-libs=../../../graph-compiler/externals/llvm-project/build/Release/lib/libmlir_runner_utils.so ../../../graph-compiler/test/mlir/test/gc/gpu-runner/mlp.mlir
Unranked Memref base@ = 0x55c191cd5180 rank = 2 offset = 0 sizes = [1, 10] strides = [10, 1] data =
[[0.1,   0.1,   0.1,   0.1,   0.1,   0.1,   0.1,   0.1,   0.1,   0.1]]
```
Does this command fails for you? What's the failure?

AndreyPavlenko:
> does that mean that I can simply ignore the error?

Got it. The error message is redundant here. It should be either replaced with a debug message or removed.

PR #416, opened 2024-12-01
[XeVM] Lowering to LLVM 

akroviakov:
This PR introduces the first step toward lowering XeVM to LLVM.
The approach is similar to Triton, but with some differences (e.g., cache control attributes).
Conversion tests are borrowed and adapted from Triton.

For now, OpenCL intrinsics are used for 2D block operations, waiting for the corresponding SPIR-V builtins to become available.


kurapov-peter:
I think if you split it into meaningful patches we can start lending them. Looks good.

PR #417, opened 2024-12-04
Fix openvino e2e python execution

zhczhong:
Build the GcCpuRuntime as a static library and link it at compile time so that the openvino python package don't need to search it at runtime


zhczhong:
There are some failed tests and I will check how to fix them

kurapov-peter:
Those failures are due to dnnl microkernel library dependency removal.

PR #418, opened 2024-12-09
[GpuOclRuntime] Add spirv dump option

kurapov-peter:
This should help with kernel debugging. Also, no need to rely on the driver's env variables we had problems with.

### DIFF
--- a/lib/gc/ExecutionEngine/GPURuntime/ocl/GpuOclRuntime.cpp
+++ b/lib/gc/ExecutionEngine/GPURuntime/ocl/GpuOclRuntime.cpp
@@ -128,10 +128,9 @@ struct Kernel {
 
   explicit Kernel(cl_program program, cl_kernel kernel, const size_t *gridSize,
                   const size_t *blockSize, size_t argNum, const size_t *argSize)
-      : program(program),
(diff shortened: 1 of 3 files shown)


lib/gc/ExecutionEngine/GPURuntime/ocl/GpuOclRuntime.cpp
```diff
@@ -993,6 +993,43 @@ OclModuleBuilder::build(const OclRuntime::Ext &ext) {
...
+    mod->walk([&](LLVM::GlobalOp global) {
+      auto isaKernel = [&](LLVM::GlobalOp op) {
+        return op.getName().starts_with("gcGpuOclKernel_") &&
+               op.getName().ends_with("SPIRV");
+      };
+
+      if (!isaKernel(global))
+        return WalkResult::skip();
+
+      auto name = global.getName();
```
AndreyPavlenko, line 1006:
```suggestion
      auto name = global.getName();
      
      if (!name.starts_with("gcGpuOclKernel_") && !name.ends_with("SPIRV") {
        return WalkResult::skip();
      }
```

I think, lambda is redundant here, it's called only once.

dchigarev (approved):

PR #422, opened 2024-12-15
[XeVM] Attach target pass

akroviakov:
This PR adds a pass that attaches an xevm target to a GPU module. This enables generating binaries similar to the upstream via `gpu-module-to-binary`.

### DIFF
--- a/include/gc/Transforms/Passes.td
+++ b/include/gc/Transforms/Passes.td
@@ -150,6 +150,41 @@ def GpuTilingAndFusion : Pass<"gpu-tiling", "func::FuncOp"> {
            "The maximum workgroup size.">
     ];
 }
+
+def GpuXeVMAttachTarget: Pass<"xevm-attach-target", ""> {
+  let summary = "Attaches a XeVM target attribute to a GPU Module.";
+  let description = [{
+    This pass searches for all GPU Modules in the immediate regions and attaches
+    a XeVM target if the module matches the name specified by the `module` argument.
+
+    Example:
+    ```
+    // File: in.mlir:
+    gpu.module @nvvm_module_1 {...}
+    gpu.module @rocdl_module_2 {...}
+    gpu.module @xevm_module_3 {...}
+    // mlir-opt --xevm-attach-target="module=xevm.* chip=pvc" in.mlir
+    gpu.module @nvvm_module_1 {...}
+    gpu.module @rocdl_module_2 {...}
+    gpu.module @xevm_module_3 [#xevm.target<chip = "pvc">] {...}
+    ```
+  }];
+  let options = [
+    Option<"moduleMatcher", "module", "std::string",
+           /*default=*/ [{""}],
+           "Regex used to identify the modules to attach the target to.">,
+    Option<"triple", "triple", "std::string",
+           /*default=*/ "\"spirv64-unknown-unknown\"",
+           "Target triple.">,
+    Option<"chip", "chip", "std::string",
+           /*default=*/"\"pvc\"",
+           "Target chip.">,
+    Option<"optLevel", "O", "unsigned",
+           /*default=*/"2",
+           "Optimization level.">
+  ];
+}
+
 #endif // GC_USE_IMEX
 
 def IterativeTilingAndFusion : Pass<"iterative-tiling-and-fusion",
--- a/lib/gc/Transforms/GPU/XeVMAttachTarget.cpp
+++ b/lib/gc/Transforms/GPU/XeVMAttachTarget.cpp
@@ -1,0 +1,70 @@
+//===-- XeVMAttachTarget.cpp - DESC -----------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements the `GpuXeVMAttachTarget` pass, attaching `#xevm.target`
+// attributes to GPU modules.
+//
+//===----------------------------------------------------------------------===//
+
+#include "gc/Dialect/LLVMIR/XeVMDialect.h"
+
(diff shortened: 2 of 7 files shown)


include/gc/Transforms/Passes.td
```diff
@@ -150,6 +150,41 @@ def GpuTilingAndFusion : Pass<"gpu-tiling", "func::FuncOp"> {
...
+    Example:
+    ```
+    // File: in.mlir:
+    gpu.module @xevm_module_1 {...}
+    gpu.module @xevm_module_2 {...}
+    gpu.module @xevm_module_1 {...}
+    // mlir-opt --xevm-attach-target="module=xevm.* chip=pvc" in.mlir
+    gpu.module @xevm_module_1 {...}
+    gpu.module @xevm_module_2 {...}
+    gpu.module @xevm_module_1 [#xevm.target<chip = "pvc">] {...}
```
kurapov-peter, line 169: This is confusing, the description says it attaches the attribute to all modules matching the name but the example doesn't follow that rule.

lib/gc/Transforms/GPU/XeVMAttachTarget.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+#include "gc/Target/LLVM/XeVM/Target.h"
+#include "gc/Transforms/Passes.h"
+
+#include "mlir/Dialect/GPU/IR/GPUDialect.h"
+#include "mlir/Dialect/GPU/Transforms/Passes.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/Pass/Pass.h"
+#include "llvm/Support/Regex.h"
+#include <iostream>
```
kurapov-peter, line 24: redundant

include/gc/Transforms/Passes.td
```diff
@@ -150,6 +150,41 @@ def GpuTilingAndFusion : Pass<"gpu-tiling", "func::FuncOp"> {
            "The maximum workgroup size.">
     ];
 }
+
+def GpuXeVMAttachTarget: Pass<"xevm-attach-target", ""> {
```
kurapov-peter, line 154: Can it be a gpu-module pass?

lib/gc/Transforms/GPU/XeVMAttachTarget.cpp
```diff
@@ -0,0 +1,82 @@
...
+  return nullptr;
+}
+
+void XeVMAttachTarget::runOnOperation() {
+  OpBuilder builder(&getContext());
+  auto target = builder.getAttr<XeVMTargetAttr>(optLevel, triple, chip);
+  llvm::Regex matcher(moduleMatcher);
+  for (Region &region : getOperation()->getRegions())
+    for (Block &block : region.getBlocks())
+      for (auto module : block.getOps<gpu::GPUModuleOp>()) {
```
kurapov-peter, line 68: This is a weird way to go through gpu modules, is there a specific reason to do this over declaring the pass to be matching gpu modules only? Is it to support the regex? Can't you check the name of the module upon matching with the same logic?

akroviakov: It follows the upstream approach:
https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/GPU/Transforms/NVVMAttachTarget.cpp#L69-L71
https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/GPU/Transforms/ROCDLAttachTarget.cpp#L77-L79

I suppose not all gpu modules may be meant to receive a target of this pass.

kurapov-peter (approved):

lib/gc/Transforms/GPU/XeVMAttachTarget.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+DictionaryAttr XeVMAttachTarget::getFlags(OpBuilder &builder) const {
+  UnitAttr unitAttr = builder.getUnitAttr();
+  SmallVector<NamedAttribute, 2> flags;
+  auto addFlag = [&](StringRef flag) {
+    flags.push_back(builder.getNamedAttr(flag, unitAttr));
+  };
+  if (!flags.empty())
+    return builder.getDictionaryAttr(flags);
+  return nullptr;
```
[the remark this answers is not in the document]
akroviakov: Yes, the upstream for different targets uses this method for flags like `ftz` or `fast`: https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/GPU/Transforms/NVVMAttachTarget.cpp#L52-L55.
For now, we do not have a set of required/exposed flags, but it is likely to change soon.

lib/gc/Transforms/GPU/XeVMAttachTarget.cpp
```diff
@@ -0,0 +1,81 @@
...
+    for (Block &block : region.getBlocks())
+      for (auto module : block.getOps<gpu::GPUModuleOp>()) {
+        // Check if the name of the module matches.
+        if (!moduleMatcher.empty() && !matcher.match(module.getName()))
+          continue;
+        // Create the target array.
+        SmallVector<Attribute> targets;
+        if (std::optional<ArrayAttr> attrs = module.getTargets())
+          targets.append(attrs->getValue().begin(), attrs->getValue().end());
+        targets.push_back(target);
```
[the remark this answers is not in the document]
akroviakov: Each module can have multiple targets that shouldn't be duplicated. Doing it via set semantics would still require us to materialize the unique targets into an array to pass ArrayRef to `getArrayAttr(targets)`, so how better would it be different to what we (and others) already have?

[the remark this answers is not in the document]
AndreyPavlenko: Sorry, I missed that the array is also created in the same loop.

lib/gc/Transforms/GPU/XeVMAttachTarget.cpp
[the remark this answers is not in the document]
akroviakov: Sure, no problem, I have addressed it in the recent commit.

PR #425, opened 2025-01-05
[XeVM] Add first integration tests

akroviakov:
This PR contains the necessary changes to launch an XeVM integration test via `gc-cpu-runner`. 

- Changes to `OpenCLRuntimeWrappers` improve error decoding and establish one static queue (this ensures no *implicit* new queue creation which turned out to be problematic).

- Load/store intrinsics only work with the  **maximum** sub-group size (not the **active** sub-group size which can be smaller), as noted in the [restrictions](https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_subgroups.html). 
For example, on `GPU Max 1100` the test runs as expected when launched with 32 threads. Although the full sub-group is still formed even if launched with 16 threads, it is not of the maximum size and hence the result is not as expected (UB according to docs). It is possible to [select](https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_required_subgroup_size.html) the maximum subgroup size:

> The optional `__attribute__((intel_reqd_sub_group_size(<int>)))` can be used to indicate that the kernel must be compiled and executed with the specified sub-group size. When this attribute is present, `get_max_sub_group_size()` is guaranteed to return the specified integer value.


lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
```diff
@@ -23,12 +23,79 @@
...
 
 namespace {
+/* clang-format off */
+#define CaseToString(x) case x: return #x;
+const char* opencl_errstr(cl_int err) {
+  switch (err){
+    CaseToString(CL_SUCCESS)
+    CaseToString(CL_DEVICE_NOT_FOUND)
+    CaseToString(CL_DEVICE_NOT_AVAILABLE)
+    CaseToString(CL_COMPILER_NOT_AVAILABLE)
```
kurapov-peter, line 33: Could you please check if this exists anywhere in ocl headers?

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
```diff
@@ -290,7 +358,14 @@ static cl_program loadModule(GPUCLQUEUE *queue, const unsigned char *data,
...
         "'-printregusage -enableBCR' -cl-kernel-arg-info -x spir";
   }
-  CL_SAFE_CALL(clBuildProgram(program, 0, NULL, build_flags, NULL, NULL));
+  err = clBuildProgram(program, 1, &queue->device_, build_flags, NULL, NULL);
+  if (err != CL_SUCCESS) {
+    char log[10240];
+    clGetProgramBuildInfo(program, queue->device_, CL_PROGRAM_BUILD_LOG,
+                          sizeof(log), log, nullptr);
+    fprintf(stderr, "Build failed: %s\n", std::string(log).c_str());
+    abort();
```
kurapov-peter, line 367: This should be recoverable I think, @AndreyPavlenko thoughts?

kurapov-peter: Also, this might be unsafe due to the hardcoded log size.

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
```diff
@@ -346,6 +421,12 @@ static void launchKernel(GPUCLQUEUE *queue, cl_kernel kernel, size_t gridX,
   }
   size_t globalSize[3] = {gridX * blockX, gridY * blockY, gridZ * blockZ};
   size_t localSize[3] = {blockX, blockY, blockZ};
+  size_t sgSize;
```
kurapov-peter, line 424: this can have a poison init value

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
```diff
@@ -447,29 +528,35 @@ extern "C" OCL_RUNTIME_EXPORT void gpuWait(GPUCLQUEUE *queue) {
...
   lastQueue = ret;
   return ret;
 }
 
+GPUCLQUEUE *getOrCreateStaticQueue() {
+  if (!lastQueue) {
+    return mgpuStreamCreate();
+  }
+  return lastQueue;
+}
```
kurapov-peter, line 544: I think we already had it somewhere? @AndreyPavlenko

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
```diff
@@ -447,29 +528,35 @@ extern "C" OCL_RUNTIME_EXPORT void gpuWait(GPUCLQUEUE *queue) {
...
-  return allocDeviceMemory(queue, size, /*alignment*/ 64, isShared);
+  return allocDeviceMemory(queue ? queue : getOrCreateStaticQueue(), size,
+                           /*alignment*/ 64, isShared);
 }
 
 extern "C" OCL_RUNTIME_EXPORT void mgpuMemFree(void *ptr, GPUCLQUEUE *queue) {
-  lastQueue = queue;
   if (ptr) {
-    deallocDeviceMemory(queue, ptr);
+    deallocDeviceMemory(queue ? queue : getOrCreateStaticQueue(), ptr);
```
kurapov-peter, line 559: this looks a bit suspicious, what's the scenario for the case when we have to create a queue to deallocate memory?

lib/gc/Target/LLVM/XeVM/Target.cpp
```diff
@@ -144,9 +144,11 @@ SpirSerializer::moduleToObject(llvm::Module &llvmModule) {
     getOperation().emitError() << "Failed translating the module to Binary.";
     return std::nullopt;
   }
-
-  StringRef bin(serializedSPIRVBinary->c_str(),
-                serializedSPIRVBinary->size() + 1);
+  if (serializedSPIRVBinary->size() % 4) {
+    getOperation().emitError() << "SPIRV code size must be a multiple of 4.";
+    return std::nullopt;
+  }
```
kurapov-peter, line 150: Is this a translator's silent failure that can result in this check being true?

src/gc-cpu-runner/CMakeLists.txt
```diff
@@ -32,5 +32,5 @@ if(GC_DEV_LINK_LLVM_DYLIB)
 endif()
 
 gc_add_mlir_tool(gc-cpu-runner gc-cpu-runner.cpp)
-target_link_libraries(gc-cpu-runner PRIVATE GcCpuRuntime)
+target_link_libraries(gc-cpu-runner PRIVATE GcJitWrapper GcCpuRuntime)
```
kurapov-peter, line 35: Is this for testing? The dependency chain should probably look differently, e.g., the runtime to depend on the wrapper if that is necessary at all. What produces the dependency?

kurapov-peter (commented):
Nice, thanks! Looks good overall, some minor comments inlined.

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
AndreyPavlenko: This function is called from MLIR and there is no logic for the error recovery.

To get the actual log size, I think, you could do the following:
```C++
size_t logLen;
clGetProgramBuildInfo(program, queue->device_, CL_PROGRAM_BUILD_LOG, 0, nullptr, &logLen);
```

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
```diff
@@ -346,6 +421,12 @@ static void launchKernel(GPUCLQUEUE *queue, cl_kernel kernel, size_t gridX,
   }
   size_t globalSize[3] = {gridX * blockX, gridY * blockY, gridZ * blockZ};
   size_t localSize[3] = {blockX, blockY, blockZ};
+  size_t sgSize;
+  CL_SAFE_CALL(clGetKernelSubGroupInfo(
+      kernel, queue->device_, CL_KERNEL_MAX_SUB_GROUP_SIZE_FOR_NDRANGE,
+      sizeof(globalSize), &globalSize, sizeof(sgSize), &sgSize, nullptr));
```
AndreyPavlenko, line 427: sgSize is never used.
Are you sure this is required? The sizes should be the same.

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
AndreyPavlenko: This is in the GpuOclRuntime.

include/gc/Dialect/LLVMIR/XeVMOps.td
```diff
@@ -221,6 +221,73 @@ def XeVM_BlockPrefetch2dOp : XeVM_Op<"blockprefetch2d">,
...
+
+      D = C + A x B
+
+      where the A, B, C input matrices and the result D have shapes:
+        D : MxN
+        C : MxN
+        A : MxK
+        B : KxN
+
+        M : repeat count, must be 1, 2, 4, or 8
```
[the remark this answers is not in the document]
akroviakov: Removed `"repeat count"` and added a link to the documentation

lib/gc/Conversion/XeVMToLLVM/XeVMToLLVM.cpp
```diff
@@ -115,13 +143,13 @@ getCacheControlMetadata(ConversionPatternRewriter &rewriter, OpType op,
...
-                         ArrayRef<std::pair<unsigned, StringRef>> paramAttrs,
-                         LLVMFuncAttributeOptions funcAttributeOptions) {
-  auto moduleOp = rewriter.getBlock()->getParent()->getParentOfType<ModuleOp>();
+static LLVM::CallOp createDeviceFunctionCall(
+    ConversionPatternRewriter &rewriter, StringRef funcName, Type retType,
+    ArrayRef<Type> argTypes, ArrayRef<Value> args,
+    mlir::ArrayRef<std::pair<unsigned, mlir::StringRef>> paramAttrs,
+    LLVMFuncAttributeOptions funcAttributeOptions) {
+  auto moduleOp =
+      rewriter.getBlock()->getParent()->getParentOfType<gpu::GPUModuleOp>();
```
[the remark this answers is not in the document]
akroviakov: Generalized to an op with a `SymbolTable` trait and added an assert.

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
akroviakov: Seems to not be the case

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
akroviakov: Generalized to a string with the size being queried as suggested

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
akroviakov: The `gpu-runner` turned out a bit problematic.
In the upstream, `gpu-to-llvm` does the type conversion including memrefs (e.g., for kernel arguments) and lowers runtime calls before running `gpu-module-to-binary` (this ensures uniformly lowered kernel arguments/parameters). However, `gpu-to-gpuocl` seems to be doing the work of `gpu-to-llvm` *after* `gpu-module-to-binary`. 
Assuming we need to lower memref for `gpu-module-to-binary` to succeed, how can the ops that consume memref in the subsequent `gpu-to-gpuocl` pass (e.g., allocOp) remain legal? 

`gpu-to-gpuocl` seems to be mainly built with regard to how IMEX handles GPUX with its related custom passes, but I'm not sure how relevant that is when we aim for the upstream flow.

lib/gc/Target/LLVM/XeVM/Target.cpp
akroviakov: This is a requirement for the resulting SPIRV to be valid.
The previous `serializedSPIRVBinary->size() + 1` was the source of `Code size must be a multiple of 4 but is X` type of error, hence the check explicitly verifies the resulting code size.

src/gc-cpu-runner/CMakeLists.txt
akroviakov: Yes it is needed for testing, like we have in 
```
target_link_libraries(GCExecutionEngineTests
  PRIVATE
  GcJitWrapper
  GcCpuRuntime
)
```
Functionally, it is needed for `gc::initCompilerAndGetDialects()` in `gc-cpu-runner.cpp`.

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
akroviakov: Code was removed

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
AndreyPavlenko: gpu-to-gpuocl converts a subset of gpu dialect ops (launch, alloc, dealloc, memcpy) to the runtime function calls.

akroviakov: Correct. The upstream `gpu-to-llvm`, however, [only legalizes](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Conversion/GPUCommon/GPUToLLVMConversion.cpp#L942-L943) the launchOp arguments, [but does not lower the op itself](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Conversion/GPUCommon/GPUToLLVMConversion.cpp#L557-L559). The launchOp [is lowered during translation to LLVM IR](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Target/LLVMIR/Dialect/GPU/GPUToLLVMIRTranslation.cpp#L55-L57) (e.g., [mlir-translate test](https://github.com/llvm/llvm-project/blob/main/mlir/test/Target/LLVMIR/gpu.mlir)).
 
Normally, the binary string is generated somewhere between `gpu-to-llvm` and translation by the `gpu-module-to-binary` pass (e.g., the [nvvm pipeline](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/GPU/Pipelines/GPUToNVVMPipeline.cpp#L93-L97)). 

Hence it is unusual from the upstream perspective to expect a binary string available before `gpu-to-llvm`.  In the case of `gpu-to-gpuocl`, both the binary and the memrefs (required due to allocOp legality) are expected to exist simultaneously. However, *in the upstream flow*, how can we generate a binary for a kernel that has memref arguments if the memrefs themselves have not yet been lowered?

test/mlir/test/gc/cpu-runner/GPU/lit.local.cfg
```diff
@@ -0,0 +1,2 @@
+if not config.gc_use_imex:
+    config.unsupported = True
```
kurapov-peter, line 2: What's missing for it to be enabled by default?

akroviakov: It seems that IMEX in gc is often a synonym for GPU ([[1]](https://github.com/intel/graph-compiler/blob/main/lib/gc/ExecutionEngine/CMakeLists.txt#L3-L6), [[2]](https://github.com/intel/graph-compiler/blob/main/lib/gc/CAPI/CMakeLists.txt#L7-L9), [[3]](https://github.com/intel/graph-compiler/blob/main/lib/gc/ExecutionEngine/Driver/CMakeLists.txt#L30-L32), [[4]](https://github.com/intel/graph-compiler/blob/main/lib/gc/Transforms/CMakeLists.txt#L43-L45)), including testing. So I follow it for now.

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
AndreyPavlenko: > The launchOp [is lowered during translation to LLVM IR](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Target/LLVMIR/Dialect/GPU/GPUToLLVMIRTranslation.cpp#L55-L57) (e.g., [mlir-translate test](https://github.com/llvm/llvm-project/blob/main/mlir/test/Target/LLVMIR/gpu.mlir)).

This is launch_func, not the launch op. According to the [doc](https://mlir.llvm.org/docs/Dialects/GPU/#gpulaunch_func-gpulaunchfuncop), gpu.launch operations are lowered to gpu.launch_func and the kernel arguments could be of type memref. So, after that, when gpu.launch ops are lowered to gpu.launch_func, the pass gpu-to-gpuocl could be run to lower these ops to the runtime function calls.

akroviakov: Yes, sorry for not being specific about the operation. 

The core issue I wanted to highlight is the unusual dependency that  `gpu-to-gpuocl` has: it expects both allocOp and the binary string simultaneously.

In the upstream flow, allocOp is only legal with memrefs, but a binary can be obtained once memrefs are lowered (uniformly in `gpu-to-llvm`).  
Hence, the `gpu-to-llvm` becomes a pre-requisite for a binary in the upstream flow. `gpu-to-gpuocl`, while having similar functionality to `gpu-to-llvm`, already expects the binary as input, but how could this binary be generated (using upstream passes)?

The argument types are not the issue, but the dependency is.

AndreyPavlenko: What about running gpu-to-gpuocl twice? First one to lower allocs and the second - to lower gpu.launch_func.

akroviakov: Yes, a "second pass" would be a solution. Maybe we could add some option flag for `gpu-to-gpuocl` to indicate the current mode (or create a new separate pass). There would be more changes needed though, for example, `gpu-to-gpuocl` relies on [querying](https://github.com/intel/graph-compiler/blob/main/lib/gc/Transforms/GPU/GpuToGpuOcl.cpp#L369) imex-specific binary attribute of a GPU module. The upstream operates on GPU objects. Required changes could be outside of the scope of this PR.

test/mlir/test/gc/cpu-runner/GPU/lit.local.cfg
kurapov-peter: It doesn't need to be if you are not using anything from imex. The only reason it appears as a synonym is that the tests are using something from imex.

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
AndreyPavlenko: Sure, we could add the required changes. I'm not sure though if the option flag is really required. On the first run there will not be gpu.launch_func ops, so only the allocs to be lowered.

akroviakov: So the integration test would need to have a `gpu.launch` instead of `gpu.launch_func` and do the following:
1. gpu-to-gpuocl to convert `gpu.alloc`, etc.
2. outlining pass to create launchFuncOp from the launchOp
3. gpu-module-to-binary
4. gpu-to-gpuocl to convert launchFuncOp and the binary to a runtime call.

Do I understand the intent correctly?

AndreyPavlenko: Perhaps, something like that.

test/mlir/test/gc/cpu-runner/GPU/lit.local.cfg
kurapov-peter: Sure, go for it

lib/gc/Target/LLVM/XeVM/Target.cpp
kurapov-peter: > The previous serializedSPIRVBinary->size() + 1 was the source

Was the size calculation incorrect or are there cases when the returned size in not a multiple of 4?

test/mlir/test/gc/cpu-runner/GPU/lit.local.cfg
akroviakov: I have decoupled IMEX, GPU components and GPU passes. 
Now one should be able to run gpu- and xevm-related passes without any special flags.
`GC_ENABLE_GPU`  :  enable `gc-gpu-runner` and runtime components (e.g., `GpuToGpuOcl.cpp`).
`GC_ENABLE_IMEX` : enable IMEX-dependent tests/passes (e.g., `LinalgToXeGPU.cpp`).

kurapov-peter (approved):

Garra1980:
Say, we want to start upstreaming xeVM dialect in current form - what exactly changes should we take from this repo and move to llvm?

akroviakov:
The upstream-relevant xevm is almost entirely covered by:
```
include/gc/Transforms/Passes.td
test/mlir/test/gc/Transforms/GPU/xevm-attach-target.mlir
test/mlir/test/gc/Transforms/GPU/module-to-binary-xevm.mlir
test/mlir/test/gc/Conversion/GPU/XeVMToLLVM/*
XeVMToLLVMIRTRanslation.cpp
XeVMToLLVM.cpp
XeVMOps.td
XeVMAttachTarget.cpp
XeVMDialect.cpp
+ relevant cmake files
```
Integration tests are not very relevant for upstreaming.

The miscellaneous changes to wrappers and `gpu-runner` are gc-specific.

lib/gc/ExecutionEngine/OpenCLRuntime/OpenCLRuntimeWrappers.cpp
akroviakov: Added `gc-gpu-runner` support for xevm tests, the support relies on outlining pass for running `gpu-to-gpuocl` twice. I will stick to the upstream flow using wrappers in the follow-up work for convenience, so it would be nice for wrapper changes to remain (no other GPU tests use them anyways).

lib/gc/Transforms/GPU/OCL/GpuToGpuOcl.cpp
```diff
@@ -246,10 +246,13 @@ struct ConvertLaunch final : ConvertOpPattern<gpu::LaunchFuncOp> {
 
     int i = 0;
     for (auto arg : kernelArgs) {
-      if (auto type = gpuLaunch.getKernelOperand(i++).getType();
+      if (auto type = gpuLaunch.getKernelOperand(i).getType();
```
AndreyPavlenko, line 249:
```suggestion
      if (auto type = gpuLaunch.getKernelOperand(i++).getType();
```

lib/gc/Transforms/GPU/OCL/GpuToGpuOcl.cpp
```diff
@@ -258,6 +261,7 @@ struct ConvertLaunch final : ConvertOpPattern<gpu::LaunchFuncOp> {
         rewriter.create<LLVM::StoreOp>(loc, arg, ptr);
         args.emplace_back(ptr);
       }
+      i++;
```
AndreyPavlenko, line 264:
```suggestion
```

lib/gc/Transforms/GPU/OCL/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+gc_add_mlir_library(GcGpuOCLPasses
```
AndreyPavlenko, line 1:
```suggestion
gc_add_mlir_library(GcGpuOclPasses
```

lib/gc/Transforms/GPU/OCL/GpuToGpuOcl.cpp
akroviakov: Addressed

akroviakov:
It seems like all of the crucial points were addressed and CI is green. Shall we merge? @AndreyPavlenko @kurapov-peter

lib/gc/Transforms/GPU/OCL/GpuToGpuOcl.cpp
```diff
@@ -352,32 +354,67 @@ struct ConvertLaunch final : ConvertOpPattern<gpu::LaunchFuncOp> {
...
 
-  // Create a new kernel and save the pointer to the global variable
-  // ...name_Ptr.
-  bool createKernel(
-      gpu::LaunchFuncOp &gpuLaunch, OpAdaptor &adaptor,
-      ConversionPatternRewriter &rewriter, const Location &loc, ModuleOp &mod,
-      StringRef funcName,
-      const std::function<SmallString<128> &(const char *chars)> &str) const {
-    auto kernelModName = gpuLaunch.getKernelModuleName();
+  StringAttr getBinaryAttrIMEX(ConversionPatternRewriter &rewriter,
```
AndreyPavlenko, line 357: Sorry, didn't notice it the last time. The methods getBinaryAttrIMEX() and getBinaryAttrUpstream() are quite similar. Let's merge them into a single one, for example getBinaryString(), and use `#ifdef GC_USE_IMEX` inside this method only to find the binary.

src/gc-gpu-runner/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@
 # SPDX-License-Identifier: Apache-2.0
 ################################################################################
 
-if(NOT GC_ENABLE_TOOLS OR NOT GC_ENABLE_IMEX)
+if(NOT GC_ENABLE_TOOLS OR NOT GC_ENABLE_GPU)
```
AndreyPavlenko, line 18:
```suggestion
if(NOT GC_ENABLE_TOOLS AND NOT GC_ENABLE_GPU AND NOT GC_ENABLE_IMEX)
```
Is it correct or I have missed something?

AndreyPavlenko (approved):

src/gc-gpu-runner/CMakeLists.txt
akroviakov: In this PR, `GC_ENABLE_IMEX` is decoupled from the general GPU tools and is used to enable IMEX-specific passes (`GcGpuIMEXPasses`) and other parts of code that rely on IMEX-specific representations.
Actually, I would assume we should not build `gpu-runner` if the tools flag is explicitly set to false. For this particular case, I followed the existing convention of allowing `gpu-runner` to be built even when we have tools disabled. What was the original intent?

AndreyPavlenko: You are right, it shouldn't be built, if the tools flag is false, but it also shouldn't be built, if neither gpu nor imex are enabled, otherwise the build will fail, I assume.

akroviakov: Then the current flag is correct: we do not build `gpu-runner` if we do not want to (tools flag is false) OR we cannot (GPU is false). This PR enables GPU-related stuff to run both with and without setting `GC_ENABLE_IMEX`, so it is no longer a pre-requisite for `gpu-runner` to work.

AndreyPavlenko: Will it work properly if tools is true, but imex and gpu are false? If yes, then it's ok.

akroviakov: Current logic:
```
if(NOT GC_ENABLE_TOOLS OR NOT GC_ENABLE_GPU)
  message(STATUS "Gpu runner is not enabled.")
  return()
endif()
```
Possible cases are:
`GC_ENABLE_TOOLS = FALSE, GC_ENABLE_GPU = FALSE` :
We do not build, because the both conditions fire.
`GC_ENABLE_TOOLS = TRUE, GC_ENABLE_GPU = FALSE `:
We do not build, because the second condition fires.
`GC_ENABLE_TOOLS = FALSE, GC_ENABLE_GPU = TRUE` :
We do not build, because the first condition fires.
`GC_ENABLE_TOOLS = TRUE, GC_ENABLE_GPU = TRUE` :
We do build, because we want AND can.

I think we can avoid possible confusion by simplifying this to:
```
if(GC_ENABLE_TOOLS AND GC_ENABLE_GPU)
  gc_add_mlir_tool(gc-gpu-runner GpuRunner.cpp)
  target_link_libraries(gc-gpu-runner PRIVATE
    GcJitWrapper
    GcGpuOclRuntime
  )
  mlir_check_all_link_libraries(gc-gpu-runner)
else
  message(STATUS "Gpu runner is not enabled.")
  return()
endif()
```

lib/gc/Transforms/GPU/OCL/GpuToGpuOcl.cpp
akroviakov: They are very similar, but we still need to locate and query different operations. I have unified both and use ifdef inside to differentiate where needed.

src/gc-gpu-runner/CMakeLists.txt
AndreyPavlenko: This will not work if GC_ENABLE_TOOLS=on and GC_ENABLE_IMEX=on.

lib/gc/Transforms/GPU/OCL/GpuToGpuOcl.cpp
```diff
@@ -360,24 +398,12 @@ struct ConvertLaunch final : ConvertOpPattern<gpu::LaunchFuncOp> {
...
-      gpuLaunch.emitOpError() << "Module " << kernelModName << " not found!";
-      return false;
-    }
-    const auto binaryAttr = kernelMod->getAttrOfType<StringAttr>("gpu.binary");
-    if (!binaryAttr) {
-      kernelMod.emitOpError() << "missing 'gpu.binary' attribute";
-      return false;
-    }
-
+    auto binaryAttr = getBinaryAttr(rewriter, gpuLaunch, kernelModName);
```
AndreyPavlenko, line 401: In case of error, empty string is returned here.

akroviakov: Addressed, now `createKernel()` returns false when `getBinaryAttr()` returns a null attribute (in case of errors).

PR #428, opened 2025-01-24
Bump LLVM

akroviakov:
Some code changes are required to shift to a fresher LLVM version.


kurapov-peter:
We could bump the imex version as well, actually. Even if it wouldn't be the same hash, there would be less changes.

Garra1980:
> We could bump the imex version as well, actually. Even if it wouldn't be the same hash, there would be less changes.

yeah, I'm actually in process of bumping llvm version for imex if this could help - https://github.com/intel-innersource/frameworks.ai.mlir.mlir-extensions/pull/925

